PR #8, opened 2020-11-20
Add Wikibase Tarball Hosting ADR

toban:
https://phabricator.wikimedia.org/T268024

decision is -> use releases.wikimedia.org

### DIFF
--- a/docs/adr/0002-tarball-hosting.md
+++ b/docs/adr/0002-tarball-hosting.md
@@ -86,4 +86,4 @@ Currently releases are being served by the Extension Distributor and the release
 * An [SRE-Access-Request] will be created, and the hike team members will be added as initial maintainers. The group will be later on expanded to cover other relevant WMDE staff.
 * Publishing step of Wikibase release pipeline will be adjusted to publish tarball release artifacts to https://releases.wikimedia.org/wikibase/ (or similar directory - final name to be defined during the implementation).
 
-[SRE-Access-Request]: https://phabricator.wikimedia.org/tag/sre-access-requests/
+[SRE-Access-Request]: https://phabricator.wikimedia.org/T268818
--- a/docs/adr/0002-tarball-hosting.md
+++ b/docs/adr/0002-tarball-hosting.md
@@ -1,10 +1,10 @@
 # 1) Use releases.wikimedia.org for hosting release tarballs {#adr_0002}
 
-Date: 2020-11-20
+Date: 2020-12-04
 
 ## Status
 
-proposed
+accepted
 
 ## Context
 
--- a/docs/adr/0002-tarball-hosting.md
+++ b/docs/adr/0002-tarball-hosting.md
@@ -53,11 +53,13 @@ Currently releases are being served by the Extension Distributor and the release
 
 #### pros: 
 
-- Seems to be the goto place for mediawiki extensions
+- Seems to be the goto place for MediaWiki extensions
 
 #### cons:
 
 - Looking through the code we would need to adapt it to serve our needs which might be undesired.
+- Does not align with our needs of simple a hosting space for artifacts.
+- Does not support other software than MediaWiki extensions (WDQS, WDQS ui etc.)
 
 ---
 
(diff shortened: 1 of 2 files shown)


docs/adr/0002-tarball-hosting.md
```diff
@@ -0,0 +1,85 @@
...
+| WF Releases   | Wikimedia          | Yes                            | Not currently          | FTP                                                  | -              | High              | Poor          | releases.wikimedia.org    |
+| Wikiba.se     | Wikimedia DE       | Yes                            | Yes                    | FTP or some kind of pulling by the server            | -              | Medium            | Poor          | wikiba.se                 |
+| ExtensionDist | Wikimedia          | Yes                            | Not currently          | Undefined, would need adaptations                    | Low (2)        | High              | Poor          | extdist.wmflabs.org/dist/ |
+
+---
+
+1. We can create new tarballs and publish without filing a ticket or speaking to admins
+2. https://grafana.wikimedia.org/d/000000161/extension-distributor-downloads?orgId=1&from=now-5y&to=now
+
+## Decision
```
toban, line 75: Should I leave the actual decision out for now until we can with certainty say releases.wikimedia.org is out of the picture?

toban: Yes

docs/adr/0002-tarball-hosting.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Extension Distributor
+
+#### pros: 
+
+- Seems to be the goto place for mediawiki extensions
+
+#### cons:
+
+- Looking through the code we would need to adapt it to serve our needs which might be undesired.
```
mariushoch, line 59: I would add that I think this it totally antagonal to what extension distributor is supposed to be: AFAIK it is a release pipeline which produces the tarball to download, while we just want a "stupid" hosting service here, were we would push our (otherwise produced) tarballs into.

docs/adr/0002-tarball-hosting.md
```diff
@@ -0,0 +1,84 @@
...
+
+---
+
+| Repository    | Service provider   | Free to WMDE                   | Self-Service (1)       | Method of publishing                                 | Current usage  | Estimated effort  | Documentation | URL
+| ------------- |------------------- | ------------------------------ | ---------------------- | ---------------------------------------------------- | -------------- | ----------------- | ------------- | --------------------------|
+| Github        | Microsoft          | Currently                      | Yes                    | Github Releases / Github Action Artifact             | -	           | Low               | Good          | github.com/wikimedia/     |
+| WF Releases   | Wikimedia          | Yes                            | Not currently          | FTP                                                  | -              | High              | Poor          | releases.wikimedia.org    |
```
mariushoch, line 66: Typo: s/WF/WMF/

docs/adr/0002-tarball-hosting.md
toban: I added another line but not sure if that matches what you was looking for, please have another look.

```
- Does not align with our needs of simple a hosting space for artifacts.

```

docs/adr/0002-tarball-hosting.md
```diff
@@ -0,0 +1,86 @@
...
+
+Date: 2020-11-20
+
+## Status
+
+proposed
+
+## Context
+
+We need to determine a place to host our new release artifacts (tarballs).
```
lucaswerkmeister, line 11: If you’re asking me to review this ADR then I need to know a bit more about what these artifacts / tarballs are. Is it one tarball per release or several? What is included in these artifacts? If there are several tarballs per release, how does the Wikibase tarball differ from what you can download from ExtensionDistributor today?

toban: Yes, sorry about just adding you without the full context.

This ADR will only concern where we are supposed to host "the release tarball(s)" (ie things the release pipeline produces)

**Is it one tarball per release or several**

This was discussed a bit yesterday and the up until then we were envisioning individual tarballs for each component(Wikibase, extensions that support it) but some good arguments for something more along the lines of a "Wikibase bundle" was presented. WDQS, WDQS-UI would be separated as they are not extensions to mediawiki. The final composition of this is still up for debate.

**If there are several tarballs per release, how does the Wikibase tarball differ from what you can download from ExtensionDistributor today**

This is one of the arguments I see for actually putting everything in one bundle/tar. Otherwise the only difference would be whats described here (https://github.com/wmde/wikibase-release-prototype/pull/10) (Will aim to bring this idea of bundling everything up for discussion again today)

manicki (approved):

tarrow (approved):

PR #10, opened 2020-11-23
Add Wikibase tarball contents ADR

toban:
https://phabricator.wikimedia.org/T267754

### DIFF
--- a/docs/adr/0003-tarball-contents.md
+++ b/docs/adr/0003-tarball-contents.md
@@ -35,22 +35,21 @@ In order to get a overview on the practices by the open-source php community a b
 
 ---
 
-Mediawiki sticks out from the crowd by including most of their release specific documentation in each of the tarball releases while all other projects in the survey do not.
-Release specific documentation seems to generally be maintained on a separate website or file that comes along side the release package.
+MediaWiki sticks out from the crowd by including most of their release specific documentation in each of the tarball releases while all other projects in the survey do not. At first glance this feels like a undesirable pattern to adopt as it's uncommon among other projects. In reality this points out a tested and proven methodology for documenting changes and shipping release notes by the mediawiki community. Release notes are added to the release branch and refined throughout the release process and then used as a template for the documentation on mediawiki.org.     
 ## Decision
 
-Seeing that release notes are generally kept separate from the actual release tarball we've decided to follow this pattern and keep release notes in a central place that allows for easy access and updates. This does not mean release notes based on the git history will not be generated automatically in the release pipeline.
+Seeing that there is an already defined process for release notes by the mediawiki community it feels natural to use existing tools and adopting this process.
 
 The `vendor` folder will not be included in the tarball as composer dependencies should be merged by the [composer merge plugin](https://github.com/wikimedia/composer-merge-plugin).  
 
 The tarball package *WILL* include a part from the source-code the following additions.
 
+- Release notes (RELEASE-NOTES-N.NN) (example RELEASE-NOTES-1.36)
+- Historic changes (HISTORY) (Previous release notes in one file)
 - git submodule dependencies
 
 The tarball will *NOT* include.
 
-- Release notes (RELEASE-NOTES-N.NN) (example RELEASE-NOTES-1.36)
-- Historic changes (HISTORY) (Previous release notes in one file)
 - `vendor` folder populated by composer
 - `build` folder
 - hidden files/folders in the root folder
--- a/docs/adr/0003-tarball-contents.md
+++ b/docs/adr/0003-tarball-contents.md
@@ -16,7 +16,7 @@ In order to get a overview on the practices by the open-source php community a b
 
 | Name      | Release notes  |   Historic release notes    | Installation instructions  | Upgrade instructions | Hidden files  | Vendor folder |
 | --------- | ---------------| --------------------------- | -------------------------- | ---------------------| --------------|---------------|
-| mediawiki |      yes       |            yes              |            yes             |          yes         |      no       |       no
+| mediawiki |      yes       |            yes              |            yes             |          yes         |      no       |       yes (9)
 | laravel   |      no (1)    |            no               |            no              |          no          |      some (6) |       no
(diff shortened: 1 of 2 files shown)


docs/adr/0003-tarball-contents.md
```diff
@@ -0,0 +1,70 @@
...
+  - `.gitreview`
+  - `.git` folder
+  - `.github` folder
+  - `.vscode` folder
+  - `.phan` folder
+  - etc.
+
+## Consequences
+
+- Do not include release notes in the tarball.
```
mariushoch, line 66: This contradicts line 46… left over form an earlier state?

docs/adr/0003-tarball-contents.md
```diff
@@ -0,0 +1,70 @@
...
+
+Seeing that there is an already defined process for release notes by the mediawiki community it feels natural to use existing tools and adopting this process.
+
+The `vendor` folder will not be included in the tarball as composer dependencies should be merged by the [composer merge plugin](https://github.com/wikimedia/composer-merge-plugin).  
+
+The tarball package *WILL* include a part from the source-code the following additions.
+
+- Release notes (RELEASE-NOTES-N.NN) (example RELEASE-NOTES-1.36)
+- Historic changes (HISTORY) (Previous release notes in one file)
+- git submodule dependencies
```
mariushoch, line 49: What does that mean? The submodules or their dependencies (however that might look)?

docs/adr/0003-tarball-contents.md
```diff
@@ -0,0 +1,70 @@
...
+---
+
+MediaWiki sticks out from the crowd by including most of their release specific documentation in each of the tarball releases while all other projects in the survey do not. At first glance this feels like a undesirable pattern to adopt as it's uncommon among other projects. In reality this points out a tested and proven methodology for documenting changes and shipping release notes by the mediawiki community. Release notes are added to the release branch and refined throughout the release process and then used as a template for the documentation on mediawiki.org.     
+## Decision
+
+Seeing that there is an already defined process for release notes by the mediawiki community it feels natural to use existing tools and adopting this process.
+
+The `vendor` folder will not be included in the tarball as composer dependencies should be merged by the [composer merge plugin](https://github.com/wikimedia/composer-merge-plugin).  
+
+The tarball package *WILL* include a part from the source-code the following additions.
```
mariushoch, line 45: s/a part/apart/

mariushoch (changes requested):

docs/adr/0003-tarball-contents.md
toban: I was referring to the submodules but did not consider their dependencies. 

I guess the dependencies of those would be either in vendor or node_modules which would be excluded?

mariushoch (approved):

mariushoch:
Sounds good to me now.

docs/adr/0003-tarball-contents.md
mariushoch: > I guess the dependencies of those would be either in vendor or node_modules which would be excluded?

I think so... and everything needed for a production setup (like minified bundle files) should be in git (that's also how we deploy this in WMF production after all).

docs/adr/0003-tarball-contents.md
```diff
@@ -0,0 +1,70 @@
...
+8. Readme.html with installation & upgrade instructions and links to documentation for further reading.
+
+---
+
+MediaWiki sticks out from the crowd by including most of their release specific documentation in each of the tarball releases while all other projects in the survey do not. At first glance this feels like a undesirable pattern to adopt as it's uncommon among other projects. In reality this points out a tested and proven methodology for documenting changes and shipping release notes by the mediawiki community. Release notes are added to the release branch and refined throughout the release process and then used as a template for the documentation on mediawiki.org.     
+## Decision
+
+Seeing that there is an already defined process for release notes by the mediawiki community it feels natural to use existing tools and adopting this process.
+
+The `vendor` folder will not be included in the tarball as composer dependencies should be merged by the [composer merge plugin](https://github.com/wikimedia/composer-merge-plugin).
```
tarrow, line 43: I'm still really unsure about this:
For example [the mediawiki installation requirements](https://www.mediawiki.org/wiki/Manual:Installation_requirements) don't include composer and [Installing mediawiki for developers](https://www.mediawiki.org/wiki/Manual:Installing_MediaWiki#Git_(for_developers)) seems to imply composer shouldn't need to be run when "just normally" installing

tarrow: [Wikibase install instructions](https://www.mediawiki.org/wiki/Wikibase/Installation) however heavily reference composer

mariushoch: From https://www.mediawiki.org/wiki/Manual:External_libraries:
> The extension distributor automatically packages composer dependencies, so tarball users won't have to do it manually.

`ExtensionProcessor::getExtraAutoloaderPaths` also seems to pick up these vendor folders… so I guess we might want to revise this? But I'm not sure how well this copes with conflicting dependency versions (I guess not well/ at all).

mariushoch: From https://www.mediawiki.org/wiki/Manual:External_libraries:
> The extension distributor automatically packages composer dependencies, so tarball users won't have to do it manually.

`ExtensionProcessor::getExtraAutoloaderPaths` also seems to pick up these vendor folders… so I guess we might want to revise this? But I'm not sure how well this copes with conflicting dependency versions (I guess not well/ at all).

toban: I poked the release team about their reasoning for including the `vendor` folder in and the arguments were the following

- Not everyone can run composer (security, shared hosting, shell access etc)
- Including it we can support both people that has access to composer and people that don't (It's easier to rm -rf vendor and the composer.lock (if you don't need them) than it is to add them later)

original task: https://phabricator.wikimedia.org/T70940

I can agree this makes sense to me. But we would probably need to update the documentation as you say @tarrow.

toban: I did a poor job at investigating the mediawiki tarballs - they indeed include the vendor.

PR #13, opened 2020-11-27
add property to item for testing

toban:
Extending the current test-case to adding a property and verifying the results a bit more in detail

### DIFF
--- a/test_scripts/selenium/queryservice-ui/queryservice-ui.page.js
+++ b/test_scripts/selenium/queryservice-ui/queryservice-ui.page.js
@@ -25,7 +25,7 @@ class QueryServiceUI extends Page {
 
 		const regexp = new RegExp( '(' + prop + ')(\\s+)(' + value + ')' );
 		const matches = text.match( regexp );
-		return matches !== null && matches.includes( prop ) && matches.includes( value );
+		return matches !== null;
 	}
 
 }
--- a/test_scripts/selenium/queryservice-ui/queryservice-ui.page.js
+++ b/test_scripts/selenium/queryservice-ui/queryservice-ui.page.js
@@ -17,6 +17,17 @@ class QueryServiceUI extends Page {
 		this.submitBtn.click();
 	}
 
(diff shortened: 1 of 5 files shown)


test_scripts/selenium/queryservice-ui/queryservice-ui.page.js
```diff
@@ -17,6 +17,17 @@ class QueryServiceUI extends Page {
...
 
+	resultIncludes( prop, value ) {
+		const text = this.resultTable.getText();
+		if ( !value ) {
+			return text.includes( prop );
+		}
+
+		const regexp = new RegExp( '(' + prop + ')(\\s+)(' + value + ')' );
+		const matches = text.match( regexp );
+		return matches !== null && matches.includes( prop ) && matches.includes( value );
```
mariushoch, line 28: The additional `String.includes` is redundant here: If the regexp matches, both `prop` and `value` are already part of `text`.

mariushoch:
Seems good to go, but giving you the chance to address the nit pick.

test_scripts/selenium/queryservice-ui/queryservice-ui.page.js
toban: It's not a string that comes back from the text.match but an array with all the groups that matched. So the prop could match but the value might not be there.

mariushoch: You are right about the array part, but still, matches will be null if only part of the string matched.

```
"foo ba".match(new RegExp( '(foo)(\\s+)(bar)' ));
null
"fo bar".match(new RegExp( '(foo)(\\s+)(bar)' ));
null
"foo bar".match(new RegExp( '(foo)(\\s+)(bar)' ));
Array(4) [ "foo bar", "foo", " ", "bar" ]
```

toban: My bad, should have confirmed that. will remove it!

PR #14, opened 2020-11-30
ADR_0005 - repository release notes process

toban:
This is just a draft for how we could do maintain release notes within the repository.

Please comment what you think about this.


docs/templates/RELEASE_NOTES-N.NN
```diff
@@ -0,0 +1,107 @@
...
+* json
+* mbstring
+* xml
+
+MySQL/MariaDB is the recommended DBMS. PostgreSQL or SQLite can also be used,
+but support for them is somewhat less mature.
+
+The supported versions are:
+
+* MySQL 5.5.8 or later
```
toban, line 93: is this true for Wikibase as well?

mariushoch: Good question… also I doubt Wikibase supports PostgreSQL. Maybe just omit this and leave it to the installation instruction?

manicki: Wikibase is only guaranteed to work on the MySQL and SQLite. Which versions I don't know, probably the closest to "confidence" we could get are the versions used on our CI builds + the MariaDB version WMF uses in production.

Postgresql I need to explicitly ask to not mention - it is not tested whatsoever, and, as of now, we don't want to make any claims about it being supported.

toban: removed this section

tarrow (approved):

PR #18, opened 2020-12-05
Add fedprops testing

toban:
Attempt at adding a way to include more test-scenarios that depending on a separate setup.  

- [X] Create an item on Wikibase and add a statement using a federated property
  - [X] verify it is available with the correct rdf through the Entity page
  - [X] verify it is available through the QueryService with the correct prefixes


Docker/LocalSettings.php.template
```diff
@@ -54,3 +54,7 @@ require_once "${DOLLAR}IP/extensions/Wikibase/repo/Wikibase.php";
 require_once "${DOLLAR}IP/extensions/Wikibase/repo/ExampleSettings.php";
 require_once "${DOLLAR}IP/extensions/Wikibase/client/WikibaseClient.php";
 require_once "${DOLLAR}IP/extensions/Wikibase/client/ExampleSettings.php";
+
+if( isset(${DOLLAR}_ENV["DOCKER_INCLUDE_SETTINGS"]) && ${DOLLAR}_ENV["DOCKER_INCLUDE_SETTINGS"] != '' ) {
```
toban, line 58: this might be a bit controversial, we could also copy this settings file (which is the one we ship in docker) and override it rather than extending it.

toban:
After mob-review:

- introduce healthchecks rather than using wait-for-it.sh
- Load all Localsettings files in a folder rather than using the env variable.

PR #33, opened 2021-01-15
Gentleman task - reorganize and rename things

toban:
- rename *_script folders
- rename build_github -> build.sh
- add --rm to publish_docker command to drop images after publishing
- move publish scripts under publish/
- move all Dockerfiles under Docker for respective stage
- add clean make targets


mariushoch:
Looks very good. Only nit pick I have is that `Docker/test/selenium/selenium/` looks a bit inconsistent (others, like `Docker/build/QueryService`, have the content in their main folder directly).

toban:
> Looks very good. Only nit pick I have is that `Docker/test/selenium/selenium/` looks a bit inconsistent (others, like `Docker/build/QueryService`, have the content in their main folder directly).

Yep that sounds like a good idea

toban:
I've re-run it about 4 times and no failures so should be good now.

PR #35, opened 2021-01-18
Fix shell linting errors

toban:
- fixes most of the shell script errors
- added a bunch of disabled rules to wait-for-it.sh as I wasn't able to make it pass the linting and still function for some reason.
- remove duplicate wait-for-it.sh script


build/build_mediawiki.sh
```diff
@@ -8,22 +8,22 @@ MEDIAWIKI_DIR="$TEMP_GIT_DIR/mediawiki"
...
 
 # remove git things from release package
 rm "$MEDIAWIKI_DIR/".git* -rfv
 
-GZIP=-9 tar -C "$TEMP_GIT_DIR" -zcf $TARBALL_PATH mediawiki
+GZIP=-9 tar -C "$TEMP_GIT_DIR" -zcf "$TARBALL_PATH" mediawiki
 
 if [ -n "$GITHUB_ENV" ]; then
-    echo "TARBALL_PATH="$TARBALL_PATH"" >> $GITHUB_ENV
+    echo "TARBALL_PATH=""$TARBALL_PATH""" >> "$GITHUB_ENV"
```
mariushoch, line 26: Does this make any difference?

build/build_queryservice.sh
```diff
@@ -10,7 +10,7 @@ wget "https://archiva.wikimedia.org/repository/releases/org/wikidata/query/rdf/s
 TARBALL_PATH="$TEMP_DIR/$TARBALL"
 
 if [ -n "$GITHUB_ENV" ]; then
-    echo "TARBALL_PATH="$TARBALL_PATH"" >> $GITHUB_ENV
+    echo "TARBALL_PATH=""$TARBALL_PATH""" >> "$GITHUB_ENV"
```
mariushoch, line 13: Does this make any difference? `echo "TARBALL_PATH="$TARBALL_PATH""` and `echo "TARBALL_PATH=""$TARBALL_PATH"""` always seem to yield the exact same result.

build/build_queryservice_ui.sh
```diff
@@ -6,29 +6,29 @@ TEMP_GIT_DIR="$(mktemp -d)"
...
 
-GZIP=-9 tar -C "$TEMP_GIT_DIR" -zcvf $TARBALL_PATH .
+GZIP=-9 tar -C "$TEMP_GIT_DIR" -zcvf "$TARBALL_PATH" .
 
-cd $ROOT
+cd "$ROOT"
 
 if [ -n "$GITHUB_ENV" ]; then
-    echo "TARBALL_PATH="$TARBALL_PATH"" >> $GITHUB_ENV
+    echo "TARBALL_PATH=""$TARBALL_PATH""" >> "$GITHUB_ENV"
```
mariushoch, line 31: Does this make any difference? `echo "TARBALL_PATH="$TARBALL_PATH""` and `echo "TARBALL_PATH=""$TARBALL_PATH"""` always seem to yield the exact same result.

build/build_wikibase.sh
```diff
@@ -26,7 +26,7 @@ GZIP=-9 tar -C "$TEMP_GIT_DIR" -zcf "$TEMP_TAR_DIR"/Wikibase.tar.gz Wikibase
 TARBALL_PATH="$TEMP_TAR_DIR/Wikibase.tar.gz"
 
 if [ -n "$GITHUB_ENV" ]; then
-    echo "TARBALL_PATH="$TARBALL_PATH"" >> $GITHUB_ENV
+    echo "TARBALL_PATH=""$TARBALL_PATH""" >> "$GITHUB_ENV"
```
mariushoch, line 29: Does this make any difference? `echo "TARBALL_PATH="$TARBALL_PATH""` and `echo "TARBALL_PATH=""$TARBALL_PATH"""` always seem to yield the exact same result.

build/write_git_metadata.sh
```diff
@@ -6,14 +6,14 @@ BUILD_METADATA_ENV_FILE=$2
...
 fi
 
-if [ -f $BUILD_METADATA_ENV_FILE ]; then
-    sed -i "/$VARIABLE_NAME/d" $BUILD_METADATA_ENV_FILE
+if [ -f "$BUILD_METADATA_ENV_FILE" ]; then
+    sed -i "/$VARIABLE_NAME/d" "$BUILD_METADATA_ENV_FILE"
 fi
 
-echo "$VARIABLE_NAME="$REPO_GIT_HASH"" >> $BUILD_METADATA_ENV_FILE
+echo "$VARIABLE_NAME=\"$REPO_GIT_HASH\"" >> "$BUILD_METADATA_ENV_FILE"
```
mariushoch, line 19: Interestingly, this will include the quotes in the output.

mariushoch:
I have some extremely perfectionist comments, mostly around the fact whether we want to have quotes around values in our env files, but other than that, this is good to go.

Regarding the environment files, for example the docker [compose manual states](https://docs.docker.com/compose/env-file/), that "There is no special handling of quotation marks. This means that they are part of the VAL.", while the [dotenv README](https://www.npmjs.com/package/dotenv) is worded differently (here surrounding quotes seem to be ignored). I don't think this matters much, but given this PR changes behaviour, I wanted to mention it (but I'm fine with ignoring it).

build/write_git_metadata.sh
tarrow: Just looking at the original code I think that might have been what was intended. After all, otherwise the final quoted section is just an empty string. I'm guessing the intention might have been to use single quotes?

build/build_wikibase.sh
tarrow: It does but it took me a while to see how. Before this commit $TARBALL_PATH is actually unquoted and followed by a quoted 0 length string.

Given how we know this command is supposed to work it mostly works exactly as planned either way. In this context spaces make no difference (they also just get echoed in like we'd expect).

An edge case where is breaks down is bash expansions. Try `TARBALL_PATH='cat dfs /*'` and you'll see is then spams in the directories of `/`

tarrow (approved):

toban:
> I have some extremely perfectionist comments, mostly around the fact whether we want to have quotes around values in our env files, but other than that, this is good to go.
> 
> Regarding the environment files, for example the docker [compose manual states](https://docs.docker.com/compose/env-file/), that "There is no special handling of quotation marks. This means that they are part of the VAL.", while the [dotenv README](https://www.npmjs.com/package/dotenv) is worded differently (here surrounding quotes seem to be ignored). I don't think this matters much, but given this PR changes behaviour, I wanted to mention it (but I'm fine with ignoring it).

lets drop the quotes for the env vars

tarrow (approved):
cool! I believe that's all the queries resolved

PR #37, opened 2021-01-19
Add queryservice cache

toban:


build/build_queryservice.sh
```diff
@@ -2,15 +2,21 @@
 set -e
 
 TARBALL="service-$QUERYSERVICE_VERSION-dist.tar.gz"
-TEMP_DIR="$(mktemp -d)"
+TARBALL_SIGNATURE="$TARBALL".md5
 
-wget "https://archiva.wikimedia.org/repository/releases/org/wikidata/query/rdf/service/$QUERYSERVICE_VERSION/$TARBALL" \
-    -O "$TEMP_DIR/$TARBALL"
+MD5_SIGNATURE="$(wget -qO- https://archiva.wikimedia.org/repository/releases/org/wikidata/query/rdf/service/$QUERYSERVICE_VERSION/$TARBALL_SIGNATURE)"
+CACHE_PATH="git_cache/queryservice/$MD5_SIGNATURE"
```
toban, line 8: maybe this should be in the cache folder rather than the git_cache folder as it's not really a git cache. thoughts?

build/build_queryservice.sh
```diff
@@ -2,15 +2,21 @@
...
 
-TARBALL_PATH="$TEMP_DIR/$TARBALL"
+if [ ! -f "$CACHE_PATH/$TARBALL" ]; then
```
toban, line 10: Might be some good arguments for moving this to the update_cache.sh script as well rather than having this logic in the  build step? Not sure.

build/build_queryservice.sh
mariushoch: Probably… the query service release is not a git repo AFAIS.

build/build_queryservice.sh
mariushoch: Not sure… it seems that `update_cache.sh` only deals with git repos.

build/build_queryservice.sh
toban: moving it there

build/build_queryservice.sh
toban: alright, added a comment and lets keep it here.

.github/workflows/built_and_test.yml
```diff
@@ -104,6 +104,14 @@ jobs:
         with:
           env-file: ${{ env.env_file }}
 
+      - name: Cache git_cache repos
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-wdqs-backend-git-repo
+        with:
+          path: git_cache
```
mariushoch, line 112: This should be `cache` now.

build/build_queryservice.sh
```diff
@@ -1,13 +1,21 @@
...
+# Downloads the service-dist.tar file from archiva and caches it 
 set -e
 
 TARBALL="service-$QUERYSERVICE_VERSION-dist.tar.gz"
-TEMP_DIR="$(mktemp -d)"
+TARBALL_SIGNATURE="$TARBALL".md5
 
-wget "https://archiva.wikimedia.org/repository/releases/org/wikidata/query/rdf/service/$QUERYSERVICE_VERSION/$TARBALL" \
-    -O "$TEMP_DIR/$TARBALL"
+MD5_SIGNATURE="$(wget -qO- "https://archiva.wikimedia.org/repository/releases/org/wikidata/query/rdf/service/$QUERYSERVICE_VERSION/$TARBALL_SIGNATURE")"
```
mariushoch, line 9: We download the md5 sum but don't use it except for assuming that it's stable per `$QUERYSERVICE_VERSION`... IMO we should either drop fetching it here (and use "wbqs-ui-$QUERYSERVICE_VERSION" in the `CACHE_PATH`) or use it to verify the tarball's integrity.

.github/workflows/built_and_test.yml
toban: doh

build/build_queryservice.sh
toban: Yeah you are right makes more sense to use the version in the path.

PR #38, opened 2021-01-19
Add repo-client testing

toban:
## Wikibase Client & Repo
- [x] Create an item on repo with sitelink to client
- [x] Create an item on repo and verify that client can reference it using wikitext
- [x] Verify that changes on repo are dispatched to client
- [ ] Create an item on repo and verify that client can use it using Lua (Scribunto)
- [ ] Create an item on repo and verify Lua cpu time is not exceeding some threshold (Scribunto)


test/docker-compose.yml
```diff
@@ -8,9 +8,9 @@ services:
       - mysql
     depends_on:
       - mysql
-    restart: unless-stopped
+    restart: always
     volumes:
-      - ./LocalSettings/${LOCALSETTINGS_VARIANT}:/var/www/html/LocalSettings
+      - ./suite-config/${LOCALSETTINGS_VARIANT}:/var/www/html/LocalSettings
```
toban, line 13: One could argue that this way of solving it has served it's purpose and should be replaced with mounting the desired template over the existing default one. Currently this supports both ways.

Docker/test/selenium/README.md
```diff
@@ -15,8 +15,9 @@
 - [x] Create an item with a statement and verify that querying using prefixes work correctly (wdt:P1 etc.. )
 
 ## Wikibase Client & Repo
-- [ ] Create an item on repo with sitelink to client and verify that changes are dispatched
-- [ ] Create an item on repo and verify that client can reference it using wikitext
+- [x] Create an item on repo with sitelink to client
+- [x] Create an item on repo and verify that client can reference it using wikitext
+- [ ] Verify that changes on repo are dispatched to client ( TODO figure out all use-cases )
```
mariushoch, line 20: What is meant with use-cases here?

Docker/test/selenium/specs/repo-client/client-item.js
```diff
@@ -0,0 +1,76 @@
...
+const WikibaseApi = require( 'wdio-wikibase/wikibase.api' );
+
+describe( 'Client Item', function () {
+
+	it( 'Special:NewItem should not be accessible on client', function () {
+
+		browser.url( process.env.MW_CLIENT_SERVER + '/wiki/Special:NewItem' );
+		$( 'h1#firstHeading' ).waitForDisplayed();
+		const notFoundText = $( 'h1#firstHeading' ).getText();
+		assert( notFoundText === 'No such special page' );
```
mariushoch, line 14: Do we really want to bind to the specific text here, or would it be better to use the message name (`?uselang=qqx`) or maybe just the http code (404)?

test/config/client/wikibase/extra-install-client.sh
```diff
@@ -0,0 +1,12 @@
...
+
+# Add sites 
+# client
+php /var/www/html/maintenance/addSite.php --conf LocalSettings.php --wiki client_wiki  --pagepath=http://wikibase.svc/wiki/\$1  --filepath=http://wikibase.svc/w/\$1 --language en --interwiki-id my_wiki my_wiki mywikigroup
+
+# Add interwiki links
+# insert interwiki link to repo on client 
+echo "INSERT INTO \`interwiki\` (\`iw_prefix\`, \`iw_url\`, \`iw_api\`, \`iw_wikiid\`, \`iw_local\`, \`iw_trans\`) VALUES ('my_wiki', 'http://wikibase.svc/wiki/\$1', 'http://wikibase.svc/w/api.php', '', '1', '0')" >> /tmp/client.sql
+
+php /var/www/html/maintenance/patchSql.php --wiki=client_wiki /tmp/client.sql
```
mariushoch, line 12:
```
echo "THE QUERY" | \
php /var/www/html/maintenance/patchSql.php --wiki=client_wiki /dev/stdin
```

should also work and doesn't require a temporary file. Maybe there's also an even simpler solution, but I haven't checked in detail.

Docker/test/selenium/README.md
toban: I thought there is a bunch of cases of changes being dispatched to the client, but I could be wrong.

The stuff I've recently touched in wikibase and are aware of are ChangeModificationNotificationJob which should show up on deletion and a change of visibility.

- ChangeVisibilityNotificationJob  Job for notifying a client wiki of a batch of revision visibility changes on the repository
- ChangeDeletionNotificationJob Job for notifying a client wiki of a batch of revision deletions on the repository.

Docker/test/selenium/specs/repo-client/client-item.js
toban: yup sounds good

test/config/client/wikibase/extra-install-client.sh
toban: Yeah sounds good!

Docker/test/selenium/README.md
toban: And I'm also not entirely sure where to draw the line here, do we want to test that all of these things work or just one single thing (which might be using the same dispatch logic)? We were discussing this during the sprint planning yesterday and there is a task coming up to try to finalize this list.

test/config/client/wikibase/extra-install-client.sh
toban: yeah no that did not work.

toban:
By the looks of it this https://phabricator.wikimedia.org/T273622 needs to get resolved before any of these tests will pass on master.

mariushoch:
Just a quick note: The new suit is not (yet) part of `make test-all`.

Docker/test/curl/test_curl.sh
```diff
@@ -1,36 +1,25 @@
...
 
-# QueryService UI test
-if curl --fail --show-error --output /dev/null --silent "$QUERYSERVICE_UI_SERVER"/; then
-    echo 'Successfully loaded the QueryService UI page!'
-else
-    echo "Could not retrieve UI root page"
-    exit 1
-fi
+    if [ -z "$URL" ]; then
+        exit 0
```
mariushoch, line 10: This seems like it should be a `return` rather… if `$URL` is empty, this script will exit successfully.

JeroenDeDauw:
This is pretty neat! Tests that run in the release/dev process are great. Will they also be runnable by people that use the published images, to verify their install is functioning properly? If not, please consider that a feature request, I think that would be really valuable.

mariushoch:
Looks fine except for the two tiny things noted above.

manicki:
> Will they also be runnable by people that use the published images, to verify their install is functioning properly? If not, please consider that a feature request, I think that would be really valuable.

@JeroenDeDauw we're thinking of something along these lines, yes.
Not something to expect to be made available next week, but we going to follow up on this. Stay tuned!

JeroenDeDauw:
Thanks @manicki

PR #41, opened 2021-01-25
add oauth

toban:
Add OAuth extension

- adds extension
- adds some documentation how to use it with quickstatements
- remove logspam in update_cache.sh

### DIFF
--- a/Docker/test/selenium/specs/repo/special-version.js
+++ b/Docker/test/selenium/specs/repo/special-version.js
@@ -1,0 +1,27 @@
+'use strict';
+
+const assert = require( 'assert' );
+
+describe( 'Special:Version', function () {
+
+	const extensions = {
+		wikibase: [ 'WikibaseClient', 'WikibaseRepository' ],
+		other: [ 'OAuth' ]
+	};
+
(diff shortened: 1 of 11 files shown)


Docker/test/selenium/specs/repo/special-version.js
```diff
@@ -0,0 +1,27 @@
...
+describe( 'Special:Version', function () {
+
+	const extensions = {
+		wikibase: [ 'WikibaseClient', 'WikibaseRepository' ],
+		other: [ 'OAuth' ]
+	};
+
+	// should be disabled for dynamic tests
+	// eslint-disable-next-line mocha/no-setup-in-describe
+	[ 'wikibase', 'other' ].forEach( ( extensionPackage ) => {
```
mariushoch, line 14: At first I didn't see the connection between this and the forEach below. Maybe using `Object.keys( extensions )` here would make it clearer that we are iterating over two levels of the same thing here? Feel free to disregard this, if this doesn't sound sensible to you.

toban: sounds reasonable!

PR #50, opened 2021-01-29
Add B&C test reporting

toban:
- Adds a JsonReporter that logs the test runs to a file by suite and stores it in TestArtifacts
- Adds a reporter step that brings failing tests up to the github UI


.github/workflows/test.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        suite: [ repo, fedprops, repo-client ]
+        suite: [ repo, fedprops ]
```
tarrow, line 20: Is this supposed to be included in this commit?

toban: no that looks wrong

.github/workflows/test.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        suite: [ repo, fedprops, repo-client ]
```
mariushoch, line 20: Seems like an accidental revert

test/test_reporter.sh
```diff
@@ -0,0 +1,13 @@
...
+set -e
+
+cd test
+
+docker-compose \
+    -f docker-compose-selenium-test.yml \
+    build
+docker-compose \
+    -f docker-compose-selenium-test.yml \
+    run --rm wikibase-selenium-test npm run "test:reporter"
```
mariushoch, line 11: Where is this (`test:reporter`) defined?

mariushoch (changes requested):
Looks good in general, but didn't test.

test/test_reporter.sh
toban: this can go!

mariushoch:
Maybe `test/test_suite.sh` should remove `./test/log/selenium/result-$SUITE.json` (or `onRunnerEnd` doesn't need to amend)? Otherwise the logs just accumulate over time.

PR #51, opened 2021-02-01
[1.T272577] add helper bash scripts for cloning and cleaning

toban:
use them on queryservice ui

### DIFF
--- a/Docker/publish/tag_git/tag_git.sh
+++ b/Docker/publish/tag_git/tag_git.sh
@@ -12,7 +12,7 @@ done
 
 if [ -z "$WIKIBASE_BRANCH_NAME" ] || \
 [ -z "$RELEASE_VERSION" ] || \
-[ -z "$QUERYSERVICE_UI_COMMIT_HASH" ] || \
+[ -z "$WDQS_UI_COMMIT_HASH" ] || \
 [ -z "$WIKIBASE_COMMIT_HASH" ] || \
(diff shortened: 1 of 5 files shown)


Docker/publish/tag_git/tag_git.sh
```diff
@@ -12,7 +12,7 @@ done
 
 if [ -z "$WIKIBASE_BRANCH_NAME" ] || \
 [ -z "$RELEASE_VERSION" ] || \
-[ -z "$QUERYSERVICE_UI_COMMIT_HASH" ] || \
+[ -z "$WDQS_UI_COMMIT_HASH" ] || \
```
tarrow, line 15: If we want to make this change we need to make corresponding changes to .env and the versions/* files I think.

toban: In this case WDQS_UI_COMMIT_HASH is read (if present) from build_metadata_wdqs_ui.env which is written at checkout which is currently set by QUERYSERVICE_UI_COMMIT_HASH. 

I think these intentionally became two different variables as we want the build_metadata file to determine what commit is tagged. However looking at it again i realize the naming should be updated to reflect that one is used for building and one is used as an artifact later from the build.

toban: added the METADATA_ prefix to make this difference more clear.

PR #52, opened 2021-02-01
[2.T272577] Build quick statements

toban:


Docker/build/QuickStatements/Dockerfile
```diff
@@ -0,0 +1,44 @@
+FROM ubuntu:xenial as fetcher
+
+RUN apt-get update && \
+    apt-get install --yes --no-install-recommends git=1:2.* ca-certificates=201* && \
+    apt-get clean && rm -rf /var/lib/apt/lists/*
+
+COPY artifacts/quickstatements.tar.gz artifacts/quickstatements.tar.gz
+RUN tar xzf artifacts/quickstatements.tar.gz
+
+FROM composer:1 as composer
```
Krinkle, line 10: Might be worth "just" installing composer the boring way from a more authortative source and without custom shell script wrappers and other indirection that bring in explonentially more dependencies and unknown third parties trust into this release process.

There's a number of different installation options available at <https://getcomposer.org/download/> varying from "download a pinned version phar" to "pipe this shell command to do it for you" (the latter is auto-updates and is more executably scary, but still less so than what this foreign container can and does do at install time and run time).

Krinkle: Note that this currently ends up using php8.0 during the `composer install` step. Unless platform versions are pinned in composer.json, this may result in unintended side-effects. Might want to use `php:7.2-alpine` for this step to avoid/improve on that.

toban: Hey,

Thanks for the comments! I talked this through with @tarrow and we decided it would make sense to try using the wikimedia composer images for any composer installations. They seem to be offering what you suggested!

https://gerrit.wikimedia.org/r/plugins/gitiles/integration/config/+/1ed2425c7245202c2ae6ce357b3602f50bc6838b/dockerfiles/composer-php72/

Krinkle: Ack, these are not production grade though. These are just for throw-away CI testing contexts where the only significant influence any of the executed code has is to leave a boolean signal stating that tests pass. It has no lasting impact on anything or credentials etc.

Krinkle: Having said that, the [way it installs](https://gerrit.wikimedia.org/r/plugins/gitiles/integration/config/+/1ed2425c7245202c2ae6ce357b3602f50bc6838b/dockerfiles/composer-scratch/Dockerfile.template) is indeed one of the recommended ways of securely installing Composer, so that's certainly a good one to consider copying in here if that suites your needs :)

PR #56, opened 2021-02-04
some fixes for building REL1_35 branch

toban:
- mw REL1_35 for some reasons contain the Vector folder so this is now
removed always and cloned.
- Fedprops auto-setup is not in REL1_35 so we can use the manual setup
to skip relying on it.

### DIFF
--- a/build/build_mediawiki.sh
+++ b/build/build_mediawiki.sh
@@ -12,12 +12,15 @@ if [ ! -d "$MEDIAWIKI_DIR" ]; then
    git clone --branch "$MEDIAWIKI_BRANCH_NAME" "$ROOT/git_cache/core.git" "$MEDIAWIKI_DIR"
 fi
 
-# Add vector from cache
-if [ ! -d "$MEDIAWIKI_DIR/skins/Vector" ]; then
-   git clone --branch "$MEDIAWIKI_BRANCH_NAME" "$ROOT/git_cache/skins/Vector.git" "$MEDIAWIKI_DIR/skins/Vector"
+# remove any existing folders
+if [ -d "$MEDIAWIKI_DIR/skins/Vector" ]; then
+    rm -rf "$MEDIAWIKI_DIR/skins/Vector"
 fi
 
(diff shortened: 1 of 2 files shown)


test/suite-config/fedprops/LocalSettings.php
```diff
@@ -7,4 +7,24 @@
 $wgDebugDumpSql  = true;
 $wgShowDBErrorBacktrace = true;
 
-$wgWBRepoSettings['federatedPropertiesEnabled'] = true;
\ No newline at end of file
+$wgWBRepoSettings['federatedPropertiesEnabled'] = true;
+$wgWBRepoSettings['federatedPropertiesSourceScriptUrl'] = 'https://www.wikidata.org/w/';
```
mariushoch, line 11: Do we want to leave a comment here that this can be removed once we drop support for 1.35?

build/build_mediawiki.sh
```diff
@@ -12,12 +12,15 @@ if [ ! -d "$MEDIAWIKI_DIR" ]; then
    git clone --branch "$MEDIAWIKI_BRANCH_NAME" "$ROOT/git_cache/core.git" "$MEDIAWIKI_DIR"
 fi
 
-# Add vector from cache
-if [ ! -d "$MEDIAWIKI_DIR/skins/Vector" ]; then
-   git clone --branch "$MEDIAWIKI_BRANCH_NAME" "$ROOT/git_cache/skins/Vector.git" "$MEDIAWIKI_DIR/skins/Vector"
+# remove any existing folders
+if [ -d "$MEDIAWIKI_DIR/skins/Vector" ]; then
```
mariushoch, line 16: No need to check for existence here, `rm -rf` doesn't fail if the directory doesn't exist.

mariushoch:
I have two nitpicks, this is fine to merge otherwise.

test/suite-config/fedprops/LocalSettings.php
toban: yeah sounds good.

build/build_mediawiki.sh
toban: correct!

PR #57, opened 2021-02-04
test improvements

toban:
- dont install node_modules if it was copied on test
- shutdown all suites

### DIFF
--- a/Docker/test/selenium/Dockerfile
+++ b/Docker/test/selenium/Dockerfile
@@ -7,4 +7,4 @@ RUN pip3 install --no-cache --upgrade pip setuptools
 
 WORKDIR /usr/src/app/
 COPY . .
-RUN npm ci
+RUN if [ ! -d "/usr/src/app/node_modules" ]; then npm ci --verbose ; fi
--- a/Docker/test/selenium/Dockerfile
+++ b/Docker/test/selenium/Dockerfile
@@ -1,6 +1,6 @@
-FROM node:12.19.1-alpine3.10
+FROM docker:latest
 
-RUN apk add --update bash chromium chromium-chromedriver alpine-sdk python3
+RUN apk add --update bash chromium chromium-chromedriver alpine-sdk python3 nodejs npm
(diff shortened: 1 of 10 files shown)


Docker/test/selenium/Dockerfile
```diff
@@ -7,4 +7,4 @@ RUN pip3 install --no-cache --upgrade pip setuptools
 
 WORKDIR /usr/src/app/
 COPY . .
-RUN npm ci
\ No newline at end of file
+RUN if [ ! -d "/usr/src/app/node_modules" ]; then npm ci --verbose ; fi
```
mariushoch, line 10: Do we not need to make sure this is up to date still?

toban: I've been commenting this out manually because I have the `node_modules` folder which gets copied and then replaced. In CI this would not occur as `node_modules` is not a part of the repo or the cache. This is only here because I don't want to comment/uncomment this all the time.

So it should only happen locally, and I don't have any other ideas than making sure we would be aware this happening or maybe post a big disclaimer that it's not gonna update dependencies on a run?

mariushoch: Hm, maybe add it to `.dockerignore`? I'm really afraid this is going to add potential for very subtle problems.

toban: Ok the commit hook felt a bit too intrusive as I would have to keep installing them on manually. How about this env variable for use in `local.env` to make this decision a bit more conscious?

toban: install dependencies locally

in local.env

```
SKIP_INSTALL_TEST_DEPENDENCIES=1
```

mariushoch:
Meh, I messed up by merging #58 into this, sorry.

I'm fine with this change, but I would propose to use a less generic name. So maybe something like `SKIP_INSTALL_SELENIUM_TEST_DEPENDENCIES` instead of `SKIP_INSTALL_TEST_DEPENDENCIES` (which might mean anything).

PR #59, opened 2021-02-08
ADR0012 - versioning scheme & branching

toban:
https://phabricator.wikimedia.org/T273889

### DIFF
--- a/docs/adr/0012-versioning.md
+++ b/docs/adr/0012-versioning.md
@@ -4,7 +4,7 @@ Date: 2021-02-08
 
 ## Status
 
-proposed
+accepted
 
 ## Context
 
--- a/docs/adr/0012-versioning.md
+++ b/docs/adr/0012-versioning.md
@@ -15,18 +15,32 @@ Both of these types of branches are maintained by the WMF and will be used by th
 As there we will still rely on the release branches and deployment branches, we don't see any real benefit by maintaining an additional set of branches for minor releases. These releases could still be maintained through git tagging the WMDE controlled repositories at the point in time a release is produced ([similar to how mediawiki is maintaining their release candidates](https://gerrit.wikimedia.org/g/mediawiki/core/+/refs/tags/1.35.0-rc.0)). 
 
 
-As for the tag itself we also need to decide on version-numbering scheme for differentiating releases where the WMF version number does not change. Seeing that MediaWiki is already using the semantic versioning for incrementing their (maintenance releases) releases in between new release branches being cut it makes sense for wikibase to adopt a similar process as we will still rely on the release branches for future releases.
+As for the tag itself we also need to decide on version-numbering scheme for differentiating releases where the WMF version number does not change. Seeing that MediaWiki is already using the semantic versioning for incrementing their (maintenance releases) releases in between new release branches being cut it could make sense for wikibase to adopt a similar process as we will still rely on the release branches for future releases.
 
-Example: The patch release following 1.35 (1.35.0) will be known as 1.35.1 
+Example: The patch release following 1.35 (1.35.0) will be known as 1.35.1
+
+However adopting the same semantic versioning for Wikibase also presents some additional risks of adding confusion to what is to be included in a release as our release cycle will not always align perfectly with the one of MediaWiki.
+
+Example:
+
+1. We release 1.35.0 which contains Mediawiki 1.35.0.
+2. We release 1.35.1 which contains Mediawiki 1.35.0 and also some additional extensions and bug fix to Quickstatements.
+3. WMF releases a 1.35.1 security patch release of Mediawiki
+4. We release 1.35.2 which contains Mediawiki 1.35.1 (and the previous changes mentioned)
+
+For this reason we probably want to adopt adopt another scheme in order to avoid this confusion. Looking at the MediaWiki naming convention for release candidates (`1.35.0-rc.0`) something along these lines might be sufficient to prevent any confusion. 
 
 ## Decision
 
 - Do not create branches for every release
-- Use the semantic versioning scheme already in place.
+- Use `-wmde.N` as a suffix to indicate version where N is incremented between releases
 
 ## Consequences
 
 - We will not create branches of every minor version number we release. Instead we will use tagging.
-- Releases in between new release branches being cut will be versioned using the semantic versioning scheme already in place. 
+- Releases in between new release branches being cut will be versioned by appending our own version number to the mediawiki version.
+
+Example: A release based on 1.35.0 would be versioned as 1.35.0-wmde.0
+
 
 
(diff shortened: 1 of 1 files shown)


docs/adr/0012-versioning.md
```diff
@@ -0,0 +1,32 @@
...
+
+As there we will still rely on the release branches and deployment branches, we don't see any real benefit by maintaining an additional set of branches for minor releases. These releases could still be maintained through git tagging the WMDE controlled repositories at the point in time a release is produced ([similar to how mediawiki is maintaining their release candidates](https://gerrit.wikimedia.org/g/mediawiki/core/+/refs/tags/1.35.0-rc.0)). 
+
+
+As for the tag itself we also need to decide on version-numbering scheme for differentiating releases where the WMF version number does not change. Seeing that MediaWiki is already using the semantic versioning for incrementing their (maintenance releases) releases in between new release branches being cut it makes sense for wikibase to adopt a similar process as we will still rely on the release branches for future releases.
+
+Example: The patch release following 1.35 (1.35.0) will be known as 1.35.1
```
tarrow, line 20: I have some concerns that this could be confusing. Imagine the following situation:

We release 1.35.0 which contains Mediawiki 1.35.0.
We release 1.35.1 which contains Mediawiki 1.35.0 and also some additional extensions and bug fix to Quickstatements.
WMF releases a 1.35.1 security patch release of Mediawiki
We release 1.35.2 which contains Mediawiki 1.35.1 (and the previous changes mentioned)

This sound to me like we might confuse ourselves let-alone users.

tarrow: My strawman proposed solution stolen from the ticket is:
We release 1.35.0-wmde0 which contains Mediawiki 1.35.0.
We release 1.35.0-wmde1 which contains Mediawiki 1.35.0 and also some additional extensions and bug fix to Quickstatements.
WMF releases a 1.35.1 security patch release of Mediawiki
We release 1.35.1-wmde0 which contains Mediawiki 1.35.1 (and the previous changes mentioned)

toban: All very valid points. I've updated the ADR with the first argument and adopted your suggestion with a slight modification.

mariushoch (approved):

toban:
Should I just bump this to accepted now?

mariushoch:
> Should I just bump this to accepted now?

IMO, go for it.

docs/adr/0012-versioning.md
```diff
@@ -0,0 +1,46 @@
...
+
+Date: 2021-02-08
+
+## Status
+
+accepted
+
+## Context
+
+WMF does during their release process create a number of branches which are used during the release process to orchestrate the different tools to produce the correct artifacts and to support the continuous deployment to production. The release branches (ex. REL1_35) are cut for MediaWiki and every extension repository with the purpose of maintaining releases and the weekly deployment branches are (ex. wmf/1.36.0-wmf.13) are used to deployments to the different [groups](https://versions.toolforge.org/).
```
manicki, line 11:
```suggestion
WMF does during their release process create a number of branches which are used during the release process to orchestrate the different tools to produce the correct artifacts and to support the continuous deployment to production. The release branches (ex. REL1_35) are cut for MediaWiki and every extension repository with the purpose of maintaining releases and the weekly deployment branches are (ex. wmf/1.36.0-wmf.13) are used to deployments to the different [deployment groups](https://versions.toolforge.org/). 
```

docs/adr/0012-versioning.md
```diff
@@ -0,0 +1,46 @@
...
+
+Both of these types of branches are maintained by the WMF and will be used by the wikibase release strategy for our major releases and deployment. However we need to decide how to maintain our own intermediate releases that will be published on releases.wikimedia.org, dockerhub and tagged on git.
+
+As there we will still rely on the release branches and deployment branches, we don't see any real benefit by maintaining an additional set of branches for minor releases. These releases could still be maintained through git tagging the WMDE controlled repositories at the point in time a release is produced ([similar to how mediawiki is maintaining their release candidates](https://gerrit.wikimedia.org/g/mediawiki/core/+/refs/tags/1.35.0-rc.0)). 
+
+
+As for the tag itself we also need to decide on version-numbering scheme for differentiating releases where the WMF version number does not change. Seeing that MediaWiki is already using the semantic versioning for incrementing their (maintenance releases) releases in between new release branches being cut it could make sense for wikibase to adopt a similar process as we will still rely on the release branches for future releases.
```
manicki, line 18: > that MediaWiki is already using the semantic versioning

I don't think it is accurate. Mediawiki does not use semantic versioning I believe. They do use three-part version numbers, but this does not imply following semantic versioning. There are breaking changes made etc, and the "major" version number has ever changed.

https://www.mediawiki.org/wiki/Version_lifecycle talks about "major releases" to refer to versions 1.35, 1.36, etc. In semver's language this  does not match, as 35, 36 would be _minor_ versions.
I didn't find MW's/WMF's official name for the last  version number part.

I am intending to make a constructive suggestion but I realize now I don't full understand the point of this second sentence here.
Is it saying that as Mediawiki already uses a version numbering scheme which increments the "small" version number when there are updates (e.g. security fixes) added to the "stable release" (e.g. 1.35), and that there are git branches made using the similar naming pattern? And that could use something similar?

docs/adr/0012-versioning.md
```diff
@@ -0,0 +1,46 @@
...
+
+Example: The patch release following 1.35 (1.35.0) will be known as 1.35.1
+
+However adopting the same semantic versioning for Wikibase also presents some additional risks of adding confusion to what is to be included in a release as our release cycle will not always align perfectly with the one of MediaWiki.
```
manicki, line 22: also here: I don't think that I got the point about the semantic versioning reference, or the point is not accurate.

manicki (commented):
Overall looks good. I am having some doubts about the mentioning of the semantic versioning, but overall this document is fine and could go in in my opinion.

docs/adr/0012-versioning.md
toban: I think this whole section was my first proposal that got shot down by @tarrow. Essentially what I suggested was lets just keep using the versioning schema already in place (semantic or not) and call our next release 1.35.0 the next one in march would then be 1.35.1

docs/adr/0012-versioning.md
```diff
@@ -40,6 +39,5 @@ For this reason we probably want to adopt adopt another scheme in order to avoid
 - We will not create branches of every minor version number we release. Instead we will use tagging.
 - Releases in between new release branches being cut will be versioned by appending our own version number to the mediawiki version.
 
-Example: A release based on 1.35.0 would be versioned as 1.35.0-wmde.0
-
-
+Example: A Wikibase release based on 1.35.0 would be versioned as 1.35.0-wmde.0
+Example: A queryservice frontend release would be versioned wmde.0, wmde.1 etc.
```
manicki, line 43: thanks for this addition!

manicki (approved):

PR #63, opened 2021-02-10
fix artifacts typo

toban:


Makefile
```diff
@@ -33,26 +33,6 @@ queryservice-ui:
 	bash update_cache.sh services
 	eval ". ./build/build_queryservice_ui.sh; bash build/build_queryservice_ui_docker.sh ${QUERYSERVICE_UI_IMAGE_NAME}"
 
-oauth:
```
mariushoch, line 36: Are these removed on purpose?

toban: they are built in the dockerfile for the bundle now

mariushoch: Ah, makes sense… in that case the definition of `all` should be changed to no longer include these.

tarrow (approved):

PR #69, opened 2021-02-17
[WIP] Remove unused tarball and add wikibase dependency to bundle

toban:


tarrow:
This would then mean that this line isn't needed right? https://github.com/wmde/wikibase-release-pipeline/blob/main/.github/workflows/built_and_test.yml#L103

Would it be equally suitable to add an extra line to the make dependencies?

toban:
> This would then mean that this line isn't needed right? https://github.com/wmde/wikibase-release-pipeline/blob/main/.github/workflows/built_and_test.yml#L103
> 
> Would it be equally suitable to add an extra line to the make dependencies?

Doh. It builds the whole thing again now because the downloaded docker image isn't straight in `artifacts/`

PR #74, opened 2021-02-19
bump composer php version

toban:
Seems we need 7.3 to build REL1_35

https://github.com/wmde/wikibase-release-pipeline/actions/runs/580831354


toban:
tried it out here -> https://github.com/wmde/wikibase-release-pipeline/runs/1933907858?check_suite_focus=true

PR #81, opened 2021-03-02
[T276075] Test pingback

toban:
Tests the pingback by intercepting the message and checking the database that it was marked as sent

### DIFF
--- a/Docker/test/selenium/specs/pingback/pingback.js
+++ b/Docker/test/selenium/specs/pingback/pingback.js
@@ -1,0 +1,38 @@
+'use strict';
+
+const assert = require( 'assert' );
+const defaultFunctions = require( '../../helpers/default-functions' );
+
+describe( 'Pingback', function () {
+
+	before( function () {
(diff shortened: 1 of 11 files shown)


Docker/test/selenium/specs/pingback/pingback.js
```diff
@@ -0,0 +1,38 @@
...
+
+	it( 'Should ping on first page request', function () {
+
+		browser.url( process.env.MW_SERVER + '/wiki/Main_Page' );
+
+		browser.pause( 5 * 1000 );
+
+		const sqlResult = browser.dockerExecute(
+			process.env.DOCKER_MYSQL_NAME,
+			'mysql --user ' + process.env.DB_USER +
```
mariushoch, line 20: This will fail if `DB_USER` contains spaces/ quotes… but given this is test only I guess we can get away with this (also below).

toban: done!

PR #82, opened 2021-03-03
[T275862.1] Add WikibaseManifest to bundle

toban:
Adds some simple tests, will add a followup for the publishing (git tagging).

There is now a compatible REL1_35 branch


mariushoch:
Tested locally, looks good.

PR #83, opened 2021-03-03
[T275216] Make publish/git just output commands

toban:


mariushoch:
We also need to update `docs/topics/publishing.md`, I think.

PR #99, opened 2021-03-12
[DNM] test publishing after testing

toban:
I didn't manage to get the whole thing running within two hours.

TODO: 

- We can't run the publish thing the way it works now in the same workflow_run as the artifacts are not made public through the api before the whole thing is finished. However we could probably just download them instead.
- Alternatively we could at the end of a run that should get published use the api to queue a job with the correct parameters (https://docs.github.com/en/rest/reference/actions#create-a-workflow-dispatch-event)


toban:
closing this as we now have the packages

PR #100, opened 2021-03-12
[T277167] Add babel and test to bundle

toban:


toban:
somethings borked

Docker/test/selenium/specs/repo/elasticsearch.js
```diff
@@ -62,7 +62,7 @@ describe( 'ElasticSearch', function () {
 		// run jobs detached
 		browser.dockerExecute(
 			process.env.DOCKER_WIKIBASE_REPO_NAME,
-			'php /var/www/html/maintenance/runJobs.php --wiki my_wiki --wait --maxjobs 10',
+			'php /var/www/html/maintenance/runJobs.php --wiki my_wiki --wait --maxjobs 20',
```
manicki, line 65: what happened that this is needed? More wikitext processing/something like this due to parsing babel boxes?

manicki (approved):
Looks great, and works as written on the cover!

Docker/test/selenium/specs/repo/elasticsearch.js
toban: I'm actually not 100% sure this fixed the issue which was consistently failing this test. But looking through Babel there is only a DeferredUpdate which i guess could be picked up by the job runner? 

The other option is that this was just timing out because of more things to index for elastic-search and I was unlucky and it was consistently timing-out on Github. 

However running this locally this test still times out sometimes so I merged https://github.com/wmde/wikibase-release-pipeline/pull/102 lets see if this makes it happen less.

PR #101, opened 2021-03-12
setting wgEnableRestAPI is not necessary

toban:


mariushoch:
Per MediaWiki `includes/DefaultSettings.php`: `@deprecated Since 1.35, defaults to true and is ignored by MediaWiki core itself. No longer functions as a setting. Will be removed in 1.36.`

PR #104, opened 2021-03-16
[T195989] Add confirm-edit and simple test

toban:


manicki (approved):
Looks good and works as expected, thanks

PR #105, opened 2021-03-17
[T277166] add WikibaseLocalMedia and tests

toban:
- Adds WikibaseLocalMedia extension to docker bundle
- Adds test for uploading and using the new datatype
- Adds MW_WG_ENABLE_UPLOADS, MW_WG_UPLOAD_DIRECTORY to dockerfile env 

If I've understood this correctly we need https://github.com/ProfessionalWiki/WikibaseLocalMedia/pull/5 first for this to pass


toban:
testing running it on REL1_35 here https://github.com/wmde/wikibase-release-pipeline/actions/runs/660721663

.env
```diff
@@ -1,7 +1,6 @@
 WIKIBASE_BRANCH_NAME=master
 MEDIAWIKI_BRANCH_NAME=master
 QUERYSERVICE_VERSION=0.3.40
-QUERYSERVICE_UI_COMMIT_HASH=e84ab35125557ff073f42ba522a684d35c288b38
```
toban, line 4: This should be building against head of master

toban:
Testing another REL1_35 build https://github.com/wmde/wikibase-release-pipeline/actions/runs/665245821

test/suite-config/repo-client/docker-compose.override.yml
```diff
@@ -33,6 +33,7 @@ services:
     restart: always
     volumes:
       - ./suite-config/repo-client/client/LocalSettings.php.template:/LocalSettings.php.template
+      - ./suite-config/repo-client/LocalSettings.php:/var/www/html/LocalSettings.d/WikibaseLocalMedia.php # disable WikibaseLocalMedia
```
mariushoch, line 36: So this is mounted only to serve as a dummy to override WikibaseLocalMedia.php? That seems to point to another problem, see above.

Docker/build/WikibaseBundle/LocalSettings.WikibaseLocalMedia.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+// https://github.com/ProfessionalWiki/WikibaseLocalMedia
+
+## WikibaseLocalMedia Configuration
+wfLoadExtension( 'WikibaseLocalMedia' );
```
mariushoch, line 6: Wont this cause problems on client installations, unless someone else does the same we did below?

Probably should be guarded by `ExtensionRegistry::getInstance()->isLoaded( 'WikibaseRepository' )`.

mariushoch (changes requested):

Docker/build/WikibaseBundle/LocalSettings.WikibaseLocalMedia.php
toban: I tried this but it did not get this to work consistently on both REL1_35 and master but lets try again and fix it otherwise.

test/suite-config/repo-client/docker-compose.override.yml
toban: Yeah this is hax but as the other comment says did try it with out any success. 

Figured client only out-of-the-box setups would a be rare sight and if the user have reached that point disabling an extension would probably be something they are able to do. But lets fix this.

toban:
another REL1_35 build https://github.com/wmde/wikibase-release-pipeline/actions/runs/675603508

Docker/build/WikibaseBundle/LocalSettings.WikibaseLocalMedia.php
toban: Ok sadly this does not work in the LocalSettings.php file. It will not be loaded at that point in time but rather be in some queue to be loaded.

I dropped the if case and excluded it in the loading of the repo-client testing instead.

toban:
yet another REL1_35 build -> https://github.com/wmde/wikibase-release-pipeline/actions/runs/722474484

mariushoch:
I just looked into the WikibaseLocalMedia a bit and it seems it also needs to be loaded on client (so that the data type can be used on pages). So it seems we should also include it on client (and if that fails, there's a bug somewhere).

toban:
> I just looked into the WikibaseLocalMedia a bit and it seems it also needs to be loaded on client (so that the data type can be used on pages). So it seems we should also include it on client (and if that fails, there's a bug somewhere).

Ok hmm, does this mean we "should" be able to use WikibaseLocalMedia on a client that does not have repo enabled? WikibaseLocalMedia should have a dependency to WikibaseClient instead?

mariushoch:
> > I just looked into the WikibaseLocalMedia a bit and it seems it also needs to be loaded on client (so that the data type can be used on pages). So it seems we should also include it on client (and if that fails, there's a bug somewhere).
> 
> Ok hmm, does this mean we "should" be able to use WikibaseLocalMedia on a client that does not have repo enabled? WikibaseLocalMedia should have a dependency to WikibaseClient instead?

Yes (although skimming the WikibaseLocalMedia code it seems the wikis need to share the same `RepoGroup` for media, which is a MediaWiki thing and has nothing to do with Wikibase repos). Given that, I'm not sure how we should best ship this… maybe just ignore the `RepoGroup` thing for now and enable it everywhere (not having it on a client might also cause problems).

toban:
> > > I just looked into the WikibaseLocalMedia a bit and it seems it also needs to be loaded on client (so that the data type can be used on pages). So it seems we should also include it on client (and if that fails, there's a bug somewhere).
> > 
> > 
> > Ok hmm, does this mean we "should" be able to use WikibaseLocalMedia on a client that does not have repo enabled? WikibaseLocalMedia should have a dependency to WikibaseClient instead?
> 
> Yes (although skimming the WikibaseLocalMedia code it seems the wikis need to share the same `RepoGroup` for media, which is a MediaWiki thing and has nothing to do with Wikibase repos). Given that, I'm not sure how we should best ship this… maybe just ignore the `RepoGroup` thing for now and enable it everywhere (not having it on a client might also cause problems).

Ok so during the daily today we had a chat with everyone on what to do about this and the decision was to include it as it currently is but add some [disclaimers](https://github.com/wmde/wikibase-release-pipeline/pull/105/commits/052299e1edb2f20d6ff8596ec2742f535501825a) regarding this not working in a client only setup.

tarrow (approved):

salipio:
> We recently finished creation of another Wikibase extension that adds a new data type (EDTF https://www.loc.gov/standards/datetime/). Created for the Luxembourg Ministry of Culture. Not announced yet but we will do so soon. Would this also be welcome in this bundle?

Hey Jeroen, we're in the final days of getting the Spring Wikibase release (incl. the Docker images) finalised, so we're past the cutoff point of including anything additional in the bundle. However, I' think it would make sense for us to consider including the EDTF datatype extension in the Fall 2021 version of the Docker bundle. Besides Luxembourg, I have heard of one other project that is interested in CIDOC-CRM within Wikibase. I've created a ticket to track this topic here: https://phabricator.wikimedia.org/T280656

JeroenDeDauw:
> I have heard of one other project that is interested in CIDOC-CRM within Wikibase

The French library?

PR #111, opened 2021-03-20
Don't set RELEASE_USER to $USER

mariushoch:
`RELEASE_USER` should be set via `local.env`.


toban (commented):
Sounds good, lets add something describing how to use it?

pipeline.md mentions variables in local.env
publishing.md talks about the tar ball step

mariushoch:
> Sounds good, lets add something describing how to use it?

Good idea, done.

PR #113, opened 2021-04-06
[T268716] Add property reference + statement test

toban:
Adds the last browser tests to https://phabricator.wikimedia.org/T268716


mariushoch:
I'm not a huge fan of all the `browser.pause`, but I also don't see a better way now.

PR #115, opened 2021-04-07
Add test for search endpoints

toban:


toban:
This https://gerrit.wikimedia.org/r/c/mediawiki/extensions/Wikibase/+/677561 should maybe make this test now green now

toban:
yay works

tarrow (approved):
Only thing we considered was that there is a small amount of instability by having this really call wikidata.org. Of course if wikidata.org is down then we have bigger problems and with suspect that ISNI is gonna be a thing for a long to to come

PR #116, opened 2021-04-07
[T279538.1] Fixes visual-editor configuration

toban:
UI wasn't visible externally before.


mariushoch:
> UI wasn't visible externally before.

What does that mean? Do you have steps to reproduce?

The VE documentation says [`You should not need to change this in most cases!`](https://www.mediawiki.org/wiki/Extension:VisualEditor#Linking_with_Parsoid).

toban:
> > UI wasn't visible externally before.
> 
> What does that mean? Do you have steps to reproduce?
> 
> The VE documentation says [`You should not need to change this in most cases!`](https://www.mediawiki.org/wiki/Extension:VisualEditor#Linking_with_Parsoid).

Yeah thats why I did not include it in the first place because I thought this was gonna resolve itself once deployed somewhere with correct hostname variables, however that was not the case. When i first setup the test-system I had to include it [here](https://github.com/toban/wb-docker/blob/main/LocalSettings/LocalSettings.php). And as i started playing with the docker-compose example it didn't work locally either so It just seems it's required if it's in a docker container.

One way of testing it could be to go to the third patch in this chain test the docker-compose example and then remove that line inside the docker-container and see what happens.

mariushoch:
> One way of testing it could be to go to the third patch in this chain test the docker-compose example and then remove that line inside the docker-container and see what happens.

I did that, but still this doesn't seem to make any difference. Do you have more detailed instructions? How do you access that wiki?

toban:
@mariushoch Should be good to go now! rebased https://github.com/wmde/wikibase-release-pipeline/pull/118 aswell

PR #117, opened 2021-04-07
[T279538.3] Add example docker compose yml

toban:
Adds example/ folder containing 

- docker-compose.yml with wikibase + mysql
- docker-compose.extra.yml with additional services
- readme + env files and more examples


example/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+The wikibase bundle comes with some additional services that can be enabled.
+
+- wdqs
+- wdqs-updater
+- wdqs-frontend
+- quickstatements
+- elasticsearch
+
+### 1. Rename the `docker-compose.extra.yml` file
```
toban, line 32: Suggest running this with `-f` makes more sense

example/template.env
```diff
@@ -0,0 +1,29 @@
+# !!! TODO REPLACE WITH ACTUAL RELEASE !!!
+# Image Configuration
+WIKIBASE_IMAGE_NAME=toanwmde/wikibase:1.35.0-wmde.0
```
toban, line 3: consider moving ports to env variables.

example/docker-compose.extra.yml
```diff
@@ -0,0 +1,102 @@
...
+        aliases:
+         - elasticsearch.svc
+    environment:
+      discovery.type: single-node
+      ES_JAVA_OPTS: "-Xms512m -Xmx512m"
+  wdqs-frontend:
+    image: "${QUERYSERVICE_UI_IMAGE_NAME}"
+    restart: always
+    ports:
+    # CONFIG - Change the 8282 here to expose the Query Service UI on a different port
```
mariushoch, line 19: Should say `8834`.

example/template.env
mariushoch: Sounds like a good idea… then again, this is just an example, right? Not sure how elaborate we want to make this?

datn: Off today's meeting: I'd say we don't want to go overboard, but we do want to offer a decently flexible example config, not just throw complexity at the users and expect them to figure it out. (Which is not an argument to make things too simple, but: external simplicity translates to under-the-hood complexity, which must then be documented)

If env-variablizing the port is too much of an edge case that's fine, but if we call it out in the file we (I) need to mention it in the docs. And if it's worth calling out I thought perhaps it'd be worth turning into a variable.

example/template.env
```diff
@@ -0,0 +1,29 @@
...
+QUERYSERVICE_IMAGE_NAME=toanwmde/wdqs:0.3.40-wmde.0
+QUERYSERVICE_UI_IMAGE_NAME=toanwmde/wdqs-frontend:wmde.0
+ELASTICSEARCH_IMAGE_NAME=toanwmde/elasticsearch:6.5.4-wmde.0
+WIKIBASE_BUNDLE_IMAGE_NAME=toanwmde/wikibase-bundle:1.35.0-wmde.0
+QUICKSTATEMENTS_IMAGE_NAME=toanwmde/quickstatements:wmde.0
+MYSQL_IMAGE_NAME=mariadb:10.3
+
+# Mediawiki Configuration
+MW_ADMIN_NAME=admin
+MW_ADMIN_PASS=password
```
mariushoch, line 13: This is to short, thus MW fails to install. Setting this to `guestguest` or something like that would work. If we want to provide an intentionally broken value here, this should probably rather be left empty and a comment added.

example/README.md
toban: done

example/template.env
toban: good catch and sorry, my bad didn't try it before changing it back to a default

example/docker-compose.extra.yml
toban: removed this whole thing as it moved to the env

example/template.env
toban: done

example/README.md
toban: done

example/docker-compose.extra.yml
```diff
@@ -0,0 +1,102 @@
...
+    ports:
+     - "${WDQS_FRONTEND_PORT}:80"
+    depends_on:
+      - wdqs-proxy
+    networks:
+      default:
+        aliases:
+         - wdqs-frontend.svc
+    environment:
+      - WIKIBASE_HOST=${WIKIBASE_HOST}
```
mariushoch, line 27: Does this also need the port? (Same for `wdqs-updater` below).

toban: I think this is fine as it's going through docker internally where WDQS_HOST is the proxy
```
    location /proxy/wikibase {
        rewrite /proxy/wikibase/(.*) /${DOLLAR}1 break;
        proxy_pass http://$WIKIBASE_HOST:80;
    }

    location /proxy/wdqs {
        rewrite /proxy/wdqs/(.*) /${DOLLAR}1 break;
        proxy_pass http://$WDQS_HOST:80;
    }
```

toban: same thing with wdqs-updater

toban: wdqs-updater "should" be fine (tm)

However looking at it I see some room for improvement at  `runUpdate.sh`

```
/wait-for-it.sh "$WIKIBASE_HOST:80" -t 300 -- \
/wait-for-it.sh "$WDQS_HOST:$WDQS_PORT" -t 300 -- \
./runUpdate.sh -h http://"$WDQS_HOST":"$WDQS_PORT" -- --wikibaseUrl "$WIKIBASE_SCHEME"://"$WIKIBASE_HOST" --conceptUri "$WIKIBASE_SCHEME"://"$WIKIBASE_HOST" --entityNamespaces "$WDQS_ENTITY_NAMESPACES"

```

the WIKIBASE_HOST variable is re-used on multiple places making it a bit quirky in different situations.

--conceptUri "$WIKIBASE_SCHEME"://"$WIKIBASE_HOST" 

Which I suppose is used for outputting ` <http://wikibase.svc/entity/Q1>`

and at the same time for -wikibaseUrl "$WIKIBASE_SCHEME"://"$WIKIBASE_HOST

toban: In the example we want the updater to use the internal wikibase.svc for updating but should output `<http://localhost/entity/Q1>`

example/docker-compose.extra.yml
```diff
@@ -0,0 +1,102 @@
...
+      - 9999
+  wdqs-proxy:
+    image: "${WDQS_PROXY_IMAGE_NAME}"
+    restart: unless-stopped
+    environment:
+      - PROXY_PASS_HOST=wdqs.svc:9999
+    depends_on:
+      - wdqs
+    ports:
+      - "8989:80"
```
toban, line 53: move to env!

toban: remove! doesn't even make sense why should the proxy be public when the traffic goes through docker?

example/docker-compose.extra.yml
toban: Ok i tried updating doesn't seem to to be all that useful since it seems we would also have set WIKIBASE_HOST to the same value as WIKIBASE_CONCEPT_URI. 

If we don't do that we get the following scenario

1. concept uri set to somethingcool.svc
2. wdqs-updater goes to http://wikibase.svc to update which makes wikibase respond with <http://wikibase.svc> in EntityData. ie making this variable pointless as we would also need to switch the WIKIBASE_HOST to be the same as the concept uri.

toban: example -> https://github.com/wmde/wikibase-release-pipeline/pull/136

tarrow: I think I wasn't understanding your concerns. The thing is that you can set the hostname of the webservice to be "coolplace.io" anyway and docker should still route it the right place. I'm not sure if that makes sense?

tarrow (approved):

PR #118, opened 2021-04-08
[T279538.2] Disable magic extra install

toban:
- disables any extra-install.sh things to run by default
- elasticsearch extra-install now creates indicies as previous wikibase-docker repo
- LocalSettings.ElasticSearch.php doesn't use envsubst
- bumps curl test timeout to 120 sec


Docker/test/selenium/specs/elasticsearch/elasticsearch.js
```diff
@@ -44,7 +44,7 @@ describe( 'ElasticSearch', function () {
 		assert( alias === itemAlias );
 	} );
 
-	it( 'Should able able to run UpdateSearchIndexConfig.php', function () {
+	it.skip( 'Should able able to run UpdateSearchIndexConfig.php', function () {
```
mariushoch, line 47: What's the plan for this? Can they be re-enabled in some other form in a follow up (maybe as an assert that it actually worked)?

Docker/test/selenium/specs/elasticsearch/elasticsearch.js
```diff
@@ -57,7 +57,7 @@ describe( 'ElasticSearch', function () {
 
 	} );
 
-	it( 'Should be able to run ForceSearchIndex.php', function () {
+	it.skip( 'Should be able to run ForceSearchIndex.php', function () {
```
mariushoch, line 60: What's the plan for this? Can they be re-enabled in some other form in a follow up... maybe by making an edit and then running the script?

mariushoch (commented):
Looks good in general, but I have some questions.

Docker/test/selenium/specs/elasticsearch/elasticsearch.js
toban: Hmm I'm not sure i had a hard time deleting them because I spent so much time on it but lets just get rid of them?

toban: https://github.com/wmde/wikibase-release-pipeline/pull/124

PR #119, opened 2021-04-09
[T267712.1] Add matrix-testing of databases

toban:
- adds testing of 'mysql:5.6', 'mariadb:10.3'
- removes the assertion in the interwiki_links test because mysql gives a warning (if this would've failed everything else following would break)
- updates the screenshot writing and removes console logging


Docker/test/selenium/wdio.conf.js
```diff
@@ -78,7 +78,15 @@ exports.config = {
...
+		const screenshotPath = browser.config.screenshotPath;
+		const folder = ( process.env.DATABASE_IMAGE_NAME + '-' + process.env.SUITE ).replace( /[^a-zA-Z_0-9]/g, '_' );
+		browser.config.screenshotPath = screenshotPath + '/' + folder;
+		try {
+			saveScreenshot( test.title );
+		} catch ( error ) {
+			console.error( 'failed writing screenshot ...' );
+			console.error( error );
+		}
+		browser.config.screenshotPath = screenshotPath;
```
mariushoch, line 90: Why do you reset the `screenshotPath` here?

toban: yeah makes no sense.

toban: Or no that actually made sense because other wise we are just gonna append the folder to each test run.

```
.
├── mariadb_10_3_elasticsearch
│   ├── mariadb_10_3_elasticsearch
│   │   ├── mariadb_10_3_elasticsearch
│   │   │   ├── mariadb_10_3_elasticsearch
│   │   │   │   ├── mariadb_10_3_elasticsearch
│   │   │   │   │   └── should-be-able-to-search-via-alias.png
│   │   │   │   └── should-be-able-to-search-case-insensitive.png
│   │   │   └── Should-be-able-to-run-ForceSearchIndex.php.png
│   │   └── Should-able-able-to-run-UpdateSearchIndexConfig.php.png
│   └── Should-be-able-to-set-alias.png
└── Should-create-an-item.png

```

mariushoch (approved):

PR #120, opened 2021-04-09
[T267712.2] Add job that outputs versions used in build

toban:
- Adds PHP_VERSION env var
- Adds workflow step versions that writes built_versions.log to metadata with the versions to be used in https://phabricator.wikimedia.org/T267712


build/requirements/build_version_requirements.py
```diff
@@ -0,0 +1,31 @@
...
+
+def getVar( varName ):
+    if varName in os.environ:
+        name = components[varName]
+        return name + " " + os.environ[varName]
+    else:
+        return 'undefined'
+
+components =	{
+  "RELEASE_MAJOR_VERSION": "Mediawiki",
```
mariushoch, line 12: MediaWiki

build/requirements/build_version_requirements.py
```diff
@@ -0,0 +1,31 @@
...
+with open('config.yml') as f:
+  data = yaml.load(f, Loader=yaml.FullLoader)
+  databases = data['jobs']['test_wikibase']['strategy']['matrix']['databaseImageName']
+
+f = open("artifacts/built_versions.log", "w+")
+
+for key in components:
+  f.write(getVar(key) + '\n')
+
+f.write('Tested databases: ' + ', '.join(databases))
```
mariushoch, line 30: Not that it matters much, but the other lines are just `component version`, but this one is `component: versions`… do we maybe want to use `component: version` everywhere?

toban: sounds good

build/requirements/build_version_requirements.py
toban: done

mariushoch (approved):

PR #125, opened 2021-04-13
[T269092,T269091] Build, test and publish wdqs-proxy

toban:


Docker/build/WDQS-proxy/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM nginx:stable-alpine
+
+COPY entrypoint.sh /entrypoint.sh
+COPY wdqs.template /etc/nginx/conf.d/wdqs.template
+
+ENV PROXY_MAX_QUERY_MILLIS=60000
```
toban, line 6: Could default the host here already. Wouldn't think alot of people would ever need to change this?

tarrow (approved):

PR #127, opened 2021-04-14
[T274266] Rename queryservice 

toban:
Renames queryservice to wdqs where appropriate. 

Intentionally left the page objects in the test suite as it looked like poop when renamed to WDQSFrontend


manicki:
Looks alright, attempted to solve the merge conflict.

manicki (approved):

manicki:
Not a complaint to the change itself as it follows the established naming convention, but the name WDQS seems questionable since we are talking about Wikibase Query Service here. No "D" to refer to!

tarrow:
> Not a complaint to the change itself as it follows the established naming convention, but the name WDQS seems questionable since we are talking about Wikibase Query Service here. No "D" to refer to!

Indeed, I guess we need to ask the Discovery Team to rename it which is is probably also the right time to let them know they are now creating a different product with multiple usecases...

manicki:
Merging to move things forward.

PR #137, opened 2021-04-20
Test upgrade from old images to newly built

toban:
- Adds testing for old published images
- adds new test-upgrade make command
- adds new test-stop make command
- adds simple api test to run between upgrade and runs this in some other contexts
- updates the documentation


Docker/test/curl/test_curl.sh
```diff
@@ -20,6 +20,12 @@ check_if_up() {
 
 check_if_up "$WIKIBASE_SERVER" "/wiki/Main_Page"
 
-check_if_up "$WDQS_SERVER" "/bigdata/namespace/wdq/sparql"
+if [ -z ${SKIP_WDQS+x} ]; then
```
mariushoch, line 23: Ugh, interesting… this is essentially an isset check? Why not just check for emptiness?

tarrow (approved):

PR #138, opened 2021-04-20
example docker: use the same names as in the image

toban:
drops the mysql part in the example


mariushoch:
It seems to me that `example/docker-compose.yml` still has:
```
      - DB_USER=${MYSQL_USER}
      - DB_PASS=${MYSQL_PASSWORD}
      - DB_NAME=${MYSQL_DATABASE}

```

PR #140, opened 2021-04-21
Use docker push v4 continue-on-error

tarrow:
Possible alternative to #139


toban:
My personal opinion. Turn it off or fix the original problem is a better solution than ignoring the problem. 

If we think this can be solved by moving the uploading to happen at some later stage without risking too many simultaneous requests, then maybe we should try that. But before doing anything I think we should consider how valuable this actually is.

tarrow:
I think you're probably right; lets abandon it for now

PR #147, opened 2021-04-22
Add some more qs tests

toban:
things that doesn't work.
- Add qualifier with datatype string maybe other datatypes do not work either.
- sitelink creation doesn't seem to work for some reason.

PR #148, opened 2021-04-22
Install dependencies no longer compatible.

toban:
- removed seemingly unused python

### DIFF
--- a/Docker/build/WDQS-frontend/Dockerfile
+++ b/Docker/build/WDQS-frontend/Dockerfile
@@ -13,8 +13,8 @@ COPY --from=fetcher /wikidata-query-gui /tmp/wikidata-query-gui
 
 WORKDIR /tmp/wikidata-query-gui
 
-# Put wdqs gui in the right place
-RUN apk --no-cache add --virtual build-dependencies ca-certificates~=20191127-r2 git~=2.24 nodejs~=12 npm~=12 jq~=1.6 python~=2.7 make~=4.2 g++~=9.3
+# install dependencies
+RUN apk update && apk --no-cache add --virtual build-dependencies ca-certificates git~=2.30 nodejs~=14 npm~=14 jq~=1.6 make~=4.3 g++~=10.2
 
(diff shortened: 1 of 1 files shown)


toban:
this is what happened -> https://github.com/wmde/wikibase-release-pipeline/pull/147/checks?check_run_id=2408261583

Docker/build/WDQS-frontend/Dockerfile
```diff
@@ -13,8 +13,8 @@ COPY --from=fetcher /wikidata-query-gui /tmp/wikidata-query-gui
 
 WORKDIR /tmp/wikidata-query-gui
 
-# Put wdqs gui in the right place
-RUN apk --no-cache add --virtual build-dependencies ca-certificates~=20191127-r2 git~=2.24 nodejs~=12 npm~=12 jq~=1.6 python~=2.7 make~=4.2 g++~=9.3
+# install dependencies
+RUN apk update && apk --no-cache add --virtual build-dependencies ca-certificates git nodejs npm jq~=1.6 make g++
```
tarrow, line 17: We should probable be consistent about this and maybe just unpin (or even bump) the ca-certificates

toban: ok, dropped the certificate but still had to bump the other stuff

tarrow: gotcha, I'm not really understanding why it seems like bumping he ca-certificates in then breaking the other dependencies. For example, why do we we have to upgrade node versions?

tarrow: Also, I'm not sure I understand the need for `apk update` ? My reading of https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md#disabling-cache is that by using `--no-cache` we still get the latest versions.

toban: grand job

tarrow (approved):

PR #149, opened 2021-04-23
T280932 fix broken search

toban:
When running bundle without MW_ELASTIC_HOST we should still be able to search


tarrow (approved):

toban:
running a REL1_35 build https://github.com/wmde/wikibase-release-pipeline/actions/runs/777303192

PR #153, opened 2021-04-26
QuickStatements: Use public API url

mariushoch:
This URL is used for both internal fetches by the QuickStatements backend and for AJAX-fetches in the frontend.
Also properly set "server" (`WB_PUBLIC_HOST_AND_PORT` is not defined anywhere).


toban:
lulz no wonder it's broke AF

PR #155, opened 2021-04-26
Remove/refactor some docs in pipeline.md

toban:


docs/topics/pipeline.md
```diff
@@ -21,13 +20,6 @@ GZIP_COMPRESSION_RATE=1
 ```
 SELENIUM_LOG_LEVEL=trace
 ```
-#### Filter test-runs by file
```
toban, line 24: already described in testing.md and this is not a recommended way of using this.

PR #157, opened 2021-04-27
Add ADR-13 on latest tag

toban:
One part of https://phabricator.wikimedia.org/T275217

### DIFF
--- a/docs/adr/0013-docker-latest-tag.md
+++ b/docs/adr/0013-docker-latest-tag.md
@@ -1,0 +1,52 @@
+# 13) Docker latest tag {#adr_0013}
+
+Date: 2021-05-27
+
+## Status
+
+proposed
+
+## Context
+
+Docker allows tags to be defined for the images in a docker repository. The default tag that is used unless anything else is specified is the ["latest"](https://docs.docker.com/engine/reference/commandline/tag/#tag-an-image-referenced-by-name) tag. This goes both for publishing and pulling images, unless anything else is specified you get "latest".
+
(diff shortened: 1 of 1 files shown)


docs/adr/0013-docker-latest-tag.md
```diff
@@ -0,0 +1,52 @@
...
+
+As per previous discussions during the hike, one downside of making everything "too" easy to get up and running is that the end-users aren't fully aware what they are running or why they would need it. Therefore producing the latest tag seems like a easy thing to skip in order to avoid these kind of accidental selections and forcing the decision to be made by the user.
+
+## Decision
+
+- Do not publish a latest tag for now
+
+## Consequences
+
+- Remove any current latest / latest-base / latest-bundle tags for all images
```
manicki, line 51: technically we don't need to remove latest-base/latest-bundle tags from the "generation 1" images right?
I agree however how the latest-xxx label on image that is several years old could be confusing over time. So removing sounds fine to me

toban: True, it's not so much a consequence but a decision probably.

datn: Not sure how relevant my opinion is here, but if my review is desired: I agree that in principle we should not offer the "latest" tag to help incautious users not ruin their production deployments. Everything should be pegged to a version on our side and therefore on the user's side.

datn (approved):

manicki (approved):

tarrow (approved):

PR #166, opened 2021-05-03
add cron for every six hours for .env

toban:
Will have to be merged to main to start cronning.

- Would currently not build master since the pipeline is set to build the upcoming release.

### DIFF
--- a/.github/workflows/built_and_test.yml
+++ b/.github/workflows/built_and_test.yml
@@ -12,7 +12,8 @@ on:
   push:
 
 env:
-  env_file: ${{ github.event.inputs.env_file ||  github.event.schedule=='0 */6 * * *' && '.env' || 'versions/wmde1.env' }}
+  env_file: ${{ github.event.inputs.env_file || 'versions/wmde1.env' }}
+
 jobs:
   build_mediawiki:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/built_and_test.yml
```diff
@@ -10,8 +12,7 @@ on:
   push:
 
 env:
-  env_file: ${{ github.event.inputs.env_file || 'versions/wmde1.env' }}
-
+  env_file: ${{ github.event.inputs.env_file ||  github.event.schedule=='0 */6 * * *' && '.env' || 'versions/wmde1.env' }}
```
tarrow, line 15: I'm not sure that `github.event.schedule` will ever be set so I think this expression will always evaluate to `versions/wmde1.env` for cron triggered jobs.

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events seems to claim that the "Webhook event payload" will be "N/A" which seems to be what this `github.event` thing is populated from (https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context)

I suggest we just add the changes in lines 4 and 5 for now. In future I think we want `.env` to be be used for both cron and "CI" for PRs

toban: Seems we have to merge this thing to find out what works with schedule its not triggering anything outside the default branch.

toban: done

tarrow (approved):

PR #168, opened 2021-05-03
env file variables should be set for tests

toban:


mariushoch:
In other jobs we also pull in `variables.env`, do we (for consistency) also want to do this here?

toban:
> In other jobs we also pull in `variables.env`, do we (for consistency) also want to do this here?

Ok i just realized this won't work because we are already sourcing the .env, .variables and local.env in the make script... doh

The reason this was added was because i thought https://github.com/wmde/wikibase-release-pipeline/blob/main/Docker/test/selenium/specs/repo/special-version.js#L7 would work but it will not.

I think this maybe should get just closed.

PR #170, opened 2021-05-04
[DNM] add hacky release note generation script

toban:
- add some docs on how to use it

first -> https://gerrit.wikimedia.org/r/c/mediawiki/tools/release/+/685008

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ test/log
 diagrams/node_modules/
 artifacts/*
 local.env
+publish/release_notes/**/__pycache__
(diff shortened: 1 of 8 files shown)


manicki:
Leaving a note for reference here: @mariushoch has raised a valid concern about the mediawiki release tools code that WMDE based some of those scripts being not explicitly licensed (implicitly it is likely GPL2 or 3 licensed).
WMDE works with the WMF to provide the license information for clarity. Then, given this repository uses non-copyleft BSD 3 clause license, WMDE will likely distribute modified release note generating scripts as a separate GPL N licensed component. This repository will refer to it in the instructions

PR #187, opened 2021-06-18
Stop building the custom MW image, use the official ones.

toban:
This commit stops building the custom image and uses the images created by https://github.com/wikimedia/mediawiki-docker/blob/master/1.35/apache/Dockerfile


addshore:
I may be blind, but where is `MEDIAWIKI_IMAGE_VERSION` actually specified?

toban:
> I may be blind, but where is `MEDIAWIKI_IMAGE_VERSION` actually specified?

https://github.com/wmde/wikibase-release-pipeline/pull/187/files#diff-de15e6c3f37252e49589ea91c1541b2e87e5d051b08b175abadf88b1dbad90cfR7

addshore:
Nice, which is in turn defined in places like https://github.com/wmde/wikibase-release-pipeline/pull/193/files#diff-5f5386b5ef47d42895b529a97f7d4d4dc1e4c825615be727c210d32f6a2e76bcR13

PR #196, opened 2021-07-27
example .env: move comments to own lines

addshore:
In some environments it appears evaluation of the `.env` can lead to errors, and the comments being included in the env vars.

https://phabricator.wikimedia.org/T286334


Ladsgroup:
Legend.
Can we have a comment about this in the file to avoid making mistakes in future as well. This bug looks very sneaky.

PR #197, opened 2021-07-27
example test: ensure all expected services are up

addshore:
This potentially would have caught https://phabricator.wikimedia.org/T286334
If this ends up being red then that's true, and https://github.com/wmde/wikibase-release-pipeline/pull/196 will need merging first


addshore:
I'm not sure why only 3 checks have run here

addshore:
Oh...

```
Invalid workflow file : .github/workflows/built_and_test.yml#L492
You have an error in your yaml syntax on line 492
```

PR #199, opened 2021-08-19
bump wdqs frontend + wikibaselocalmedia versions

toban:
updating some versions for new release


Docker/build/Wikibase/Dockerfile
```diff
@@ -4,10 +4,7 @@ ARG MEDIAWIKI_IMAGE_NAME
 ARG MEDIAWIKI_IMAGE_VERSION
 FROM ubuntu:xenial as unpacker
 
-RUN apt-get update && \
```
toban, line 7: we aren't curling, unzipping or jq'ing so this can go. However this could probably just be done in the actual container it's moved to directly.

Docker/build/Wikibase/Dockerfile
```diff
@@ -27,7 +24,7 @@ RUN composer install --verbose -n --no-dev
 FROM ${MEDIAWIKI_IMAGE_NAME}:${MEDIAWIKI_IMAGE_VERSION}
 
 RUN apt-get update && \
-    DEBIAN_FRONTEND=noninteractive apt-get install --yes --no-install-recommends libbz2-dev=1.* gettext-base=0.19.* && \
```
toban, line 30: This wasn't available for installation anymore in the 1.35.3 images. https://github.com/wmde/wikibase-release-pipeline/runs/3372428965#step:8:1447

But installs as 0.21-4 https://github.com/wmde/wikibase-release-pipeline/pull/199/checks?check_run_id=3372604332#step:8:1194

lucaswerkmeister: I wonder if there was a reason to pin this at some point… I think I very vaguely remember issues with `gettext-base` in 2017 or 2018 (this would’ve been before the release pipeline, so maybe in CI?). But I think the main thing we use from that package is `envsubst`, which doesn’t sound like a hugely complicated program, so unpinning is probably okay.

lucaswerkmeister (approved):

PR #200, opened 2021-08-20
Add wikibase edtf and allow versions/.env files to support custom extension selection

toban:


lucaswerkmeister:
> This is a requirement in order to be able to do future 1.35.N releases using this same release pipeline.

Why is this a requirement? The EDTF extension claims to be compatible with MediaWiki 1.35, as far as I can tell.

build/external_extension/WikibaseLocalMedia.sh
```diff
@@ -0,0 +1,11 @@
...
+set -e
+
+## WikibaseLocalMedia
+## NOTE: WikibaseLocalMedia does currently not work in a client only setup.
+UPDATE_SUBMODULE=0 bash build/clone_repo.sh \
+    "$WIKIBASELOCALMEDIA_COMMIT_HASH" \
+    "git_cache/WikibaseLocalMedia.git" \
+    WIKIBASELOCALMEDIA \
+    "$BUILT_EXTENSIONS_PATH/WikibaseLocalMedia" \
+    master && bash build/clean_repo.sh "$BUILT_EXTENSIONS_PATH/WikibaseLocalMedia"
```
lucaswerkmeister, line 11: (Doesn’t have to be changed now, since it was this way in the old code, but) I find it pretty confusing how this has one command split across multiple lines and then at the end a second command on the same line.

update_cache.sh
```diff
@@ -57,6 +57,10 @@ for arg in "$@"; do
 		clone_if_not_present WikibaseLocalMedia.git https://github.com/ProfessionalWiki/WikibaseLocalMedia.git
 		fetch_all WikibaseLocalMedia.git
 
+		# WikibaseEdtf
+		clone_if_not_present WikibaseEdtf.git https://github.com/ProfessionalWiki/WikibaseEdtf
```
lucaswerkmeister, line 61: (super minor nitpick) I’d prefer a .git at the end of the URL just like above

lucaswerkmeister (approved):
Looks sensible to me, though I don’t entirely understand why it’s necessary.

toban:
> > This is a requirement in order to be able to do future 1.35.N releases using this same release pipeline.
> 
> Why is this a requirement? The EDTF extension claims to be compatible with MediaWiki 1.35, as far as I can tell.

I've understood it as the security release 1.35.3 we don't want to include WikibaseEdtf but make that available in the fall release first. However, the security release will happen before that (has technically already been cut) and might happen again before fall maybe.

update_cache.sh
toban: done

build/external_extension/WikibaseLocalMedia.sh
toban: done

lucaswerkmeister:
> > > This is a requirement in order to be able to do future 1.35.N releases using this same release pipeline.
> > 
> > 
> > Why is this a requirement? The EDTF extension claims to be compatible with MediaWiki 1.35, as far as I can tell.
> 
> I've understood it as the security release 1.35.3 we don't want to include WikibaseEdtf but make that available in the fall release first. However, the security release will happen before that (has technically already been cut) and might happen again before fall maybe.

Okay, that makes sense.

lucaswerkmeister (approved):

PR #205, opened 2021-08-23
Add some notes on test-system

toban:
Adds some notes on the undocumented test-system.


mariushoch:
The process seems to be somewhat hacked together, but this documentation is spot on.

PR #206, opened 2021-08-24
[T269449] Update docs, add visualization of pipeline and process

toban:
This updates the docs to make everything a bit more clear and uses the diagram part of the repo to generate a few diagrams to help with the explanations.


.github/workflows/lint.yml
```diff
@@ -37,4 +37,26 @@ jobs:
...
+          key: npm-lint-cache
+      - name: lint the selenium javascript tests
+        run: |
+          cd diagrams
+          npm install
+          npm run lint
+      - name: test building diagram
+        run: |
+          cd diagrams
+          node make_overview.js
```
toban, line 61: Could be cool to diff the existing one with what is generated here to get an indication when this should be updated. but might be le overkill.

README.md
```diff
@@ -1,5 +1,15 @@
 # Wikibase release pipeline
 
+## Repository overview
+
+The wikibase release pipeline contains a of scripts used for building, testing and publishing Wikibase docker images and tarballs.
```
addshore, line 5: a of scripts?

.github/workflows/lint.yml
addshore: This sounds exciting and like it wouldn't be too much work?

addshore (changes requested):

.github/workflows/lint.yml
toban: Ugh, not really gave it a shot yesterday and there are several annoying things.

- Id gets changed every time you rebuild so need to swap that out.
- Widths and heights seem to change every time
- It's just one line so diff isn't to happy, need to do some md5 comparison or get wdiff which isnt' installed by default.

PR #207, opened 2021-08-25
Add job runner to example

toban:
This essentially steals the entrypoint from https://gerrit.wikimedia.org/r/plugins/gitiles/releng/dev-images/+/refs/heads/master/common/jobrunner/entrypoint.sh

The repository itself does not have any LICENSE that I could find. Is this cool or do we need to re-write this tiny thing completely?

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -43,26 +43,23 @@ The wikibase bundle comes with some additional services that can be enabled.
 - quickstatements
 - elasticsearch
 
-### 1. Uncomment the extra-install scripts in `docker-compose.yml` 
-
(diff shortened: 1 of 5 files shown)


addshore:
> Okay, but I meant adding it to the [on-wiki documentation](https://www.mediawiki.org/wiki/Wikibase/Docker), not the README here :D do we expect end users to look at this README?

AFAIK currently we will only update the on wiki documentation when we officially make the next release.
And that would relate to the various components that are released.
We do not technically release or really support this docker-compose example, as in most cases (other than throw away testing) it doesn't make sense to use out of the box.

This example is currently referred to in user facing docs in ways such as:

```
You'll need a docker-compose.yml file much like the one provided in the example directory
```

PR #210, opened 2021-08-27
Update reporter to show skipped tests

toban:
If tests are skipped they are currently not shown in the report, which
makes it trickier to determine what actually happened.

This adds skips tests and shows them at the end.


toban:
https://github.com/wmde/wikibase-release-pipeline/pull/210/checks?check_run_id=3443411114#step:5:41

PR #219, opened 2021-11-12
Run linter before test happens

toban:
This is super quick and makes waiting for typos less painful.

### DIFF
--- a/test/test_suite.sh
+++ b/test/test_suite.sh
@@ -1,6 +1,13 @@
 #!/bin/bash
 set -e
 
+if [ -n "$PRE_INSTALL_SELENIUM_TEST_LINTING" ]; then
+    echo "Running javascript linter because PRE_INSTALL_SELENIUM_TEST_LINTING is set"
+    cd Docker/test/selenium/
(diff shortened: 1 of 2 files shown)


test/test_suite.sh
```diff
@@ -1,6 +1,13 @@
 #!/bin/bash
 set -e
 
+if [ -n "$PRE_INSTALL_SELENIUM_TEST_LINTING" ]; then
+    echo "Running javascript linter because PRE_INSTALL_SELENIUM_TEST_LINTING is set"
+    cd Docker/test/selenium/
+    npx eslint .
```
mariushoch, line 7: Hm, this assumes npx (npm) to be installed (but that usually only happens in selenium test containers). I personally, for example, don't have npm on my host… but I also don't see another nice way to do this, so maybe just comment that the linting will happen with npx.

addshore: the `fresh` container could also be used https://github.com/wikimedia/fresh

rti:
superseeded by #444

PR #223, opened 2021-11-12
[DNM][WIP]Make testing faster

toban:
There are steps we could take to drastically improve the speed of running the tests


toban:
This is missing some kind of check if the rebuild / loading of images is required (maybe check/store hashes of the zips), if that condition is not met this could all be skipped.

addshore:
Closing these 2 DNMs

PR #229, opened 2021-11-18
Bump wdqs to latest service release

toban:


addshore:
One thing to confirm when we make the release of the new WDQS code would be to ensure that either 1) it works with the old JNL files or 2) We document the upgrade process

PR #237, opened 2021-11-19
Update wikibase suit components versions

rosalieper:


manicki (approved):
those indeed are the most recent changes in the respective code bases

PR #249, opened 2021-11-24
Quickstatements still uses internal host

toban:
wikibase service needs to expose wikibase-docker.svc for internal calls


toban:
Can't set the default to localhost here, but adding a default internal host would allow users to set the WIKIBASE_HOST to be the external one

PR #259, opened 2021-12-01
Run against example

toban:
This 

- tweaks the example a bit and adds one missing env variable.
- fixes some breaking tests or makes them a bit more resilient (they were failing on the example)
- add a Makefile command for running tests on the example

PR #261, opened 2021-12-07
Stop writing INSTALLED_EXTENSIONS to wikibase bundle images

toban:
This method of finding out what extensions were installed was deprecated
in https://github.com/wmde/wikibase-release-pipeline/pull/259 and we can
now stop writing this in the wikibase bundle docker images.


rosalieper:
looks good to me.

PR #264, opened 2021-12-08
Add wdqs to upgrade test

toban:
This adds support for specifying a  `WDQS_SOURCE_IMAGE_NAME` when running the upgrade tests which if defined would get used in the pre-upgrade step and then later upgraded to the version built by the pipeline.

This also refactors some of the things in the upgrade test

- environment variables wasn't defined properly when creating the LocalSettings.php file
- The full log of the upgrades wasn't captured, this is now done into three different files `pre`, `post` `upgrade`

This wants to get squash-n-merged!

PR #266, opened 2021-12-14
Bump WDQS to 0.3.97 for wmde.4

toban:
this PR is based on `add-wdqs-to-upgrade-test` which would like to get merged first.


toban:
We could also just remove the streaming-updater-producer when we build here, might be the better solution.

manicki:
> We could also just remove the streaming-updater-producer when we build here, might be the better solution.

I didn't fully get this comment. Do you mean the the new version of "WDQS" includes the service ("streaming-updater-producer") packed in, that Wikibase users wouldn't use (for now)? And suggest to remove it from the resulting release packages?

toban:
> > We could also just remove the streaming-updater-producer when we build here, might be the better solution.
> 
> I didn't fully get this comment. Do you mean the the new version of "WDQS" includes the service ("streaming-updater-producer") packed in, that Wikibase users wouldn't use (for now)? And suggest to remove it from the resulting release packages?

Exactly, not sure if it is as easy as just removing it. But if it is then.

1. We don't have to worry about anyone using that thing (we don't support it AFAIK or has told anyone how to do it)
2. I contains a vulnerable version of log4j that would require a fix if used.

manicki:
> Exactly, not sure if it is as easy as just removing it. But if it is then.
> 
>     1. We don't have to worry about anyone using that thing (we don't support it AFAIK or has told anyone how to do it)
> 
>     2. I contains a vulnerable version of log4j that would require a fix if used.
> 
> 
> `streaming-updater-producer-0.3.96-jar-with-dependencies.jar`is included in the service dist

thanks. Right, it might be tempting, but that said it might lead to WMDE essentially building a custom version of WDQS, which I am not sure is what WMDE should do.
My Java days are are a distant memory now, but looking at [WDQS's pom.xml](https://github.com/wikimedia/wikidata-query-rdf/blob/444ff1b7ba11074f9ec8f07e9a334848ff5ae541/pom.xml#L808) `streaming-updater-producer` is a dependency of the whole thing. And while it does not seem to be needed or used in any of the components we expect people to use, it might still require to rebuild jars that runs without it?

Re log4j issue - if not tempering the pom file, then we might at least ensure the updater if ever used by a curious explorer, the no lookups options would be set by default. But that will probably (?) mean adding it to the example config file, which is not ideal (to reference not used components there). Ideally there would be WDQS version already up which patches the vulnerability but that might still take a bit we wouldn't like to wait.

toban:
> > Exactly, not sure if it is as easy as just removing it. But if it is then.
> > ```
> > 1. We don't have to worry about anyone using that thing (we don't support it AFAIK or has told anyone how to do it)
> > 
> > 2. I contains a vulnerable version of log4j that would require a fix if used.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `streaming-updater-producer-0.3.96-jar-with-dependencies.jar`is included in the service dist
> 
> thanks. Right, it might be tempting, but that said it might lead to WMDE essentially building a custom version of WDQS, which I am not sure is what WMDE should do. My Java days are are a distant memory now, but looking at [WDQS's pom.xml](https://github.com/wikimedia/wikidata-query-rdf/blob/444ff1b7ba11074f9ec8f07e9a334848ff5ae541/pom.xml#L808) `streaming-updater-producer` is a dependency of the whole thing. And while it does not seem to be needed or used in any of the components we expect people to use, it might still require to rebuild jars that runs without it?
> 
> Re log4j issue - if not tempering the pom file, then we might at least ensure the updater if ever used by a curious explorer, the no lookups options would be set by default. But that will probably (?) mean adding it to the example config file, which is not ideal (to reference not used components there). Ideally there would be WDQS version already up which patches the vulnerability but that might still take a bit we wouldn't like to wait.

Yeah ok, lets not tamper with the contents.

toban:
Ok this seems to pass now but lets merge https://github.com/wmde/wikibase-release-pipeline/pull/264 first

manicki:
I'd be eager to approve it once https://github.com/wmde/wikibase-release-pipeline/pull/264 has landed on the main branch,

PR #267, opened 2021-12-14
Hardcode formatMsgNoLookups=true in ES docker images

toban:


example/docker-compose.extra.yml
```diff
@@ -29,7 +29,7 @@ services:
          - elasticsearch.svc
     environment:
       discovery.type: single-node
-      ES_JAVA_OPTS: "-Xms512m -Xmx512m -Dlog4j2.formatMsgNoLookups=true"
+      ES_JAVA_OPTS: "-Xms512m -Xmx512m"
```
toban, line 32: After sleeping on it, lets keep this in because some people might still be running versions where this isn't in the container.

manicki: I agree. Out of abundance of caution, we shall keep the setting in, even if it is redundant for the "usual" use case (i.e. running the container version which already has the setting on,

toban:
> Before this gets merged we probably wanna try running this together with the setting being specified in here too (two places, what happens?) [6b1342e](https://github.com/wmde/wikibase-release-pipeline/commit/6b1342e94b1d75df1035d87d20c6e1eff47c340e)

PR #273, opened 2021-12-17
Updated mailing list URL, added more channels

manicki:
Announcement actually go to three channels


manicki:
Not sure if we can meaningfully link the Telegram channel

toban (approved):

PR #275, opened 2021-12-17
Fix the extension name for WikibaseEdtf and SyntaxHighlight in tests

rosalieper:
Bug: T297289


toban (approved):
Nice, looking at the report this seems to make them run now! Awesome! :pray:

PR #276, opened 2021-12-17
Updated bastion host name in publishing instructions

manicki:
Per data in https://wikitech.wikimedia.org/wiki/SRE/Production_access#SSH_configuration


manicki:
When I tried publishing with bast 3004, ssh connection got stuck. Maybe the host has changed and 3004 is no longer available at all?

toban (approved):
LGTM

PR #279, opened 2021-12-17
Add missing upgrade tests

toban:
Adds the missing ones and removes some that probably doesn't mean anything to anyone.

### DIFF
--- a/.github/workflows/built_and_test.yml
+++ b/.github/workflows/built_and_test.yml
@@ -493,10 +493,10 @@ jobs:
           '1.33-base',
           '1.34-base',
           '1.35-base',
-          'latest-base',
-          'lts-base',
           'wmde.0',
-          'wmde.1'
+          'wmde.1',
+          'wmde.2',
+          'wmde.3'
         ]
     needs:
       - build_wikibase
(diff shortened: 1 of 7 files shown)


manicki:
Additions look good to me. I am not sure about dropping the legacy versions already. It is still imaginable a person using the legacy docker would attempt installing the recent "new" version right? The upgrade tests would ensure that it works?
I believe we might only forget about the legacy (pre 2021) container images when the third major version of the "new" ones is published.

toban:
> > Additions look good to me. I am not sure about dropping the legacy versions already. It is still imaginable a person using the legacy docker would attempt installing the recent "new" version right? The upgrade tests would ensure that it works? I believe we might only forget about the legacy (pre 2021) container images when the third major version of the "new" ones is published.
> 
> I had a look at the checksum of those images and
> 
>     * 1.35
> 
>     * lts-base
> 
>     * latest-base
> 
>     * latest
> 
> 
> all seem to point to the same image, so we are just saving resources here.
> 
> sha256-b9f1c96ffac1166c9d64d78d751e71b86ecaf2db2e00f4cbcac922549fc195c5

And that one is still covered under 1.35-base https://hub.docker.com/layers/wikibase/wikibase/1.35-base/images/sha256-b9f1c96ffac1166c9d64d78d751e71b86ecaf2db2e00f4cbcac922549fc195c5?context=explore

manicki:
oh, okay. It is all different names for one thing. Make sense to have on test for it not multiple then. Thanks for explaining

manicki (approved):

PR #284, opened 2021-12-21
Don't upload sarif files if dependabot or forked PR

toban:


manicki (approved):

.github/actions/scan-image/action.yml
```diff
@@ -20,6 +20,9 @@ runs:
 
       # uploads it to the ui
       - uses: github/codeql-action/upload-sarif@v1
+        if: |
+          github.actor != 'dependabot[bot]' ||
```
JeroenDeDauw, line 24: Looks like this should be `&&`

toban: Awesome, thanks for that!

manicki (approved):

PR #285, opened 2021-12-21
Add MEDIAWIKI_SETTINGS_TEMPLATE_FILE

toban:
This allows for the versions to specify an LocalSettings.php.template
file that should be used when building.

I thought this was the nicest way to allow solve T259530.


test/test_upgrade.sh
```diff
@@ -55,7 +55,7 @@ TMP_LOCALSETTINGS="$TMP_DIR/LocalSettings.php"
 
 # Source the default env vars used for building and create a new LocalSettings file 
 set -o allexport; source ../Docker/build/Wikibase/default.env; set +o allexport
-envsubst < ../Docker/build/Wikibase/LocalSettings.php.template > "$TMP_LOCALSETTINGS"
+envsubst < "../Docker/build/Wikibase/$MEDIAWIKI_DOCKER_TEMPLATE_FILE" > "$TMP_LOCALSETTINGS"
```
toban, line 58: Looking at this again, this will now default to the newest template which is not what we want.

This could be resolved by either 

1. re-specify the template in the `old-versions/version.env` file 
2. we source the correct `versions/wmde.N` in the test upgrade somehow (this would only work for our releases but could be overridden with an env var in the old-version file)

toban: heh, looking at it another time this is what we want. this only happens when running against the built version which already specifies this value. nevermind! :facepalm:

toban:
I tried running this locally on wmde.3 and it works but lets run it on GH actions too https://github.com/wmde/wikibase-release-pipeline/actions/runs/1653289284

manicki (approved):
Looks good to me.
I would have probably think of it more as of a MEDIAWIKI_CONFIG_TEMPLATE_FILE, or MEDIAWIKI_SETTINGS_TEMPLATE_FILE but it is a minor details

toban:
> Looks good to me. I would have probably think of it more as of a MEDIAWIKI_CONFIG_TEMPLATE_FILE, or MEDIAWIKI_SETTINGS_TEMPLATE_FILE but it is a minor details

Yeah I Agree, `MEDIAWIKI_SETTINGS_TEMPLATE_FILE` makes more sense!

PR #288, opened 2022-01-04
adjusted description of password requirements

DavidFichtmueller:
previous description was phrased as an error message that would be shown to the user after they entered a weak password


toban:
thank you :pray:

PR #293, opened 2022-01-05
Make the names of the wikibase instance and WDQS adjustable in the settings

DavidFichtmueller:
These commits make the names of the wikibase instance and the query service adjustable via the .env file and the names are then properly displayed in Wikibase (added to the LocalSettings), Quickstatements (name of the wikibase to edit) and the query service (which is displayed in the front-end).

I created these changes for a project I am working on and thought they are useful and generic enough to benefit everybody who is installing a new wikibase. Please let me know what you think.

### DIFF
--- a/Docker/build/QuickStatements/config.json
+++ b/Docker/build/QuickStatements/config.json
@@ -4,6 +4,7 @@
 	"logfile" : "/var/log/quickstatements/tool.log" ,
 	"sites" : {
 		"${MW_SITE_NAME}" : {
+			"label" : "${MW_SITE_NAME}",
 			"oauth" : {
 				"language":"${MW_SITE_LANG}" ,
 				"project":"${MW_SITE_NAME}" ,
--- a/Docker/build/WDQS-frontend/Dockerfile
+++ b/Docker/build/WDQS-frontend/Dockerfile
@@ -36,7 +36,6 @@ COPY custom-config.json /templates/custom-config.json
 COPY default.conf /templates/default.conf
 
 ENV LANGUAGE=en\
-    BRAND_TITLE=DockerWikibaseQueryService\
     COPYRIGHT_URL=undefined
 
 ENTRYPOINT ["/entrypoint.sh"]
--- a/Docker/build/WDQS-frontend/custom-config.json
+++ b/Docker/build/WDQS-frontend/custom-config.json
@@ -15,7 +15,7 @@
       "urlShortener": "tinyurl"
     },
     "brand": {
-      "title": "$BRAND_TITLE",
+      "title": "${WDQS_NAME}",
       "logo": "logo.svg",
       "favicon": "favicon.ico",
       "copyrightUrl": "$COPYRIGHT_URL",
--- a/example/docker-compose.yml
+++ b/example/docker-compose.yml
@@ -41,6 +41,7 @@ services:
       <<: *wikibase_variables
       WIKIBASE_PINGBACK:
       MW_WG_ENABLE_UPLOADS:
+      MW_SITE_NAME: ${MW_SITE_NAME}
 
   wikibase_jobrunner:
     image: "${WIKIBASE_BUNDLE_IMAGE_NAME}"
--- a/example/docker-compose.extra.yml
+++ b/example/docker-compose.extra.yml
@@ -115,6 +115,7 @@ services:
       - "WB_ITEM_PREFIX=Item:"
       - OAUTH_CONSUMER_KEY=${OAUTH_CONSUMER_KEY}
       - OAUTH_CONSUMER_SECRET=${OAUTH_CONSUMER_SECRET}
+      - MW_SITE_NAME=${MW_SITE_NAME}
 
 volumes:
   LocalSettings:
--- a/example/docker-compose.extra.yml
+++ b/example/docker-compose.extra.yml
@@ -45,6 +45,7 @@ services:
     environment:
       - WIKIBASE_HOST=${WIKIBASE_HOST}
       - WDQS_HOST=wdqs-proxy.svc
+      - WDQS_NAME=${WDQS_NAME}
   wdqs:
     image: "${WDQS_IMAGE_NAME}"
     restart: unless-stopped
--- a/example/template.env
+++ b/example/template.env
@@ -22,6 +22,7 @@ MW_ADMIN_NAME=admin
 MW_ADMIN_EMAIL=admin@example.com
 MW_SECRET_KEY=some-secret-key
 MW_WG_ENABLE_UPLOADS=false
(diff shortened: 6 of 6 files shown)


Docker/build/QuickStatements/config.json
```diff
@@ -4,6 +4,7 @@
 	"logfile" : "/var/log/quickstatements/tool.log" ,
 	"sites" : {
 		"${MW_SITE_NAME}" : {
+			"label" : "${MW_SITE_NAME}",
```
toban, line 7: Not really sure I understand this change, isn't $MW_SITE_NAME in this file a couple of times already?

Docker/build/WDQS-frontend/custom-config.json
```diff
@@ -15,7 +15,7 @@
       "urlShortener": "tinyurl"
     },
     "brand": {
-      "title": "$BRAND_TITLE",
+      "title": "${WDQS_NAME}",
```
toban, line 18: I agree $BRAND_TITLE isn't a great choice for the env var here, but would probably maybe require updating some documentation too if you search for $BRAND_TITLE in the repo it occurs a couple of times.

Docker/build/WDQS-frontend/Dockerfile
```diff
@@ -36,7 +36,6 @@ COPY custom-config.json /templates/custom-config.json
 COPY default.conf /templates/default.conf
 
 ENV LANGUAGE=en\
-    BRAND_TITLE=DockerWikibaseQueryService\
```
toban, line 39: Shoudn't this just be renamed?

example/docker-compose.yml
```diff
@@ -41,6 +41,7 @@ services:
       <<: *wikibase_variables
       WIKIBASE_PINGBACK:
       MW_WG_ENABLE_UPLOADS:
+      MW_SITE_NAME: ${MW_SITE_NAME}
```
toban, line 44: ${MW_SITE_NAME} can be left out

toban (commented):
Hey, thanks for the PR I left some comments but will probably not have that much time to go back and forth going forward.

toban:
And this should be confirmed by building locally and running tests, or on the fork. We aren't running the pipeline on external PR:s which isn't ideal.

Docker/build/WDQS-frontend/custom-config.json
JeroenDeDauw: I thought BRAND_TITLE was clear enough and find it clearer than WDQS_NAME. Note how the JSON calls it "brand title" as well, so there also is the consistency aspect.

toban: Yeah on a second glance I agree that consistency should have priority here, lets not do this change.

DavidFichtmueller: The reason I changed this is to have all of the user adjustable settings in .env file start with WDQS_. While this brakes a bit of consistency within this template file, it increases the consistency for the user in the .env file, where they would have BRAND_TITLE next to WDQS_FRONTEND_HOST and WDQS_FRONTEND_PORT. Maybe this could be renamed to WDQS_FRONTEND_NAME though.

Docker/build/WDQS-frontend/Dockerfile
DavidFichtmueller: Maybe. I had an issue overwriting this ENV while, when I set up the configuration, and so I removed it to avoid possible confusion, with having the default "DockerWikibaseQueryService" in two different places, here and in the .env file.

Docker/build/QuickStatements/config.json
DavidFichtmueller: The ${MW_SITE_NAME} in sites is the id for this particular configuration and the one in site: defines the default configuration. These two need to be the same, though it doesn't matter what they are. The could as well just be called "default". However the one in "label" is the one that is actually displayed in the user interface. See https://quickstatements.toolforge.org/config.json for another example.

DavidFichtmueller:
I build this locally on the server I run my wikibase instance on and it works there. But of course it would be nice to have somebody else confirm this.

Docker/build/WDQS-frontend/custom-config.json
toban: Fair point, maybe a compromise could be `WDQS_FRONTEND_BRAND_TITLE`, just rolls off the tongue ;)

DavidFichtmueller: Sounds good. Now that we found a compromise, the only piece of documentation that needs to be adjusted is the [README.md](https://github.com/wmde/wikibase-release-pipeline/blob/10f338e04ff796b59e638d2a5abcebaa8e0e6d7d/Docker/build/WDQS-frontend/README.md?plain=1#L10). I can add another commit to this PR, but not before next week.

addshore:
I'm almost ready to try and merge this pull request, but I need to resolve https://github.com/wmde/wikibase-release-pipeline/pull/317 first so that CI actually runs for this PR.

Docker/build/QuickStatements/config.json
DavidFichtmueller: I noticed another issue regarding this change with Quickstatements: When a name with capital letters (or leading/tailing white spaces) is used, QS has issues, because it handles the cleanup of the names different in different parts of the code. I made a PR about it: magnusmanske/quickstatements#30 . Until this is resolved we either might have to wait or add anothers sed replacement line in the QS Docker file: https://github.com/wmde/wikibase-release-pipeline/blob/main/Docker/build/QuickStatements/Dockerfile#L35 

```sed -i "s/$site = strtolower ( trim ( get_request ( 'site' , '' ) ) ) ;/$site = get_request ( 'site' , '' ) ;/g" /var/www/html/quickstatements/public_html/api.php```

DavidFichtmueller:
See my other comment/reference PR above, this might either need to wait until the PR in QS is merged or require further adjustements as a workaround. Updating this to the latest main shouldn't be a problem. Also I need to adjust some of the variable names as we agreed in the discussion. I will do this, but probably not this or next week.

addshore:
> See my other comment/reference PR above, this might either need to wait until the PR in QS is merged or require further adjustements as a workaround.

Right you are, I'll wait until being poked again then!

addshore:
Hey @DavidFichtmueller, just checking, whats the status of this?

DavidFichtmueller:
Now that magnusmanske/quickstatements#30 has been merged, this is no longer blocking. However I currently don't have the resources to rebase this to the current main release. Maybe somebody else wants to take care of this, if not I might be able get onto this in January (earliest).

PR #299, opened 2022-01-12
Add comment on MW_ELASTIC_HOST

toban:
MW_ELASTIC_HOST gets checked at runtime if it's set in order to switch between Elasticsearch / Wikibase Search. Add a comment in the template to make this more clear.

### DIFF
--- a/example/template.env
+++ b/example/template.env
@@ -47,5 +47,7 @@ QUICKSTATEMENTS_HOST=quickstatements.svc
 QUICKSTATEMENTS_PORT=8840
 
 ## ElasticSearch
+## Comment out MW_ELASTIC_HOST to disable ElasticsSearch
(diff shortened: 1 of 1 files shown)


example/template.env
```diff
@@ -47,5 +47,7 @@ QUICKSTATEMENTS_HOST=quickstatements.svc
 QUICKSTATEMENTS_PORT=8840
 
 ## ElasticSearch
+## Comment out MW_ELASTIC_HOST to disable ElasticsSearch
+## See https://github.com/wmde/wikibase-release-pipeline/blob/main/Docker/build/WikibaseBundle/LocalSettings.d.template/WikibaseCirrusSearch.php#L6
```
manicki, line 51: That's cool. Alternatively one could also consider linking to the file and the line in a specific state of the repository (commit), so that the link certainly always point to what is meant here

manicki (approved):

PR #307, opened 2022-02-10
Share settings between jobrunner and mediawiki in a new way

addshore:
This avoids mounting all of the mediawiki & wikibase code into
a volume that will not get updated with upgrades.
Instead option to have the wikibse entry point copy LS.php into
a shared volume, specifically for the docker-compose example if
the directory exists.

Bug: T298632

### DIFF
--- a/docs/topics/upgrading.md
+++ b/docs/topics/upgrading.md
@@ -107,10 +107,10 @@ Example output
 
 ```sh
 $ docker inspect -f '{{ .Mounts }}' 916ac3ce384e
-[ {volume example_shared /var/lib/docker/volumes/example_shared/_data /var/www/html local rw true } ]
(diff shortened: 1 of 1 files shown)


addshore:
I believe this will fail CI, as this will not work until the entrypoint in a future release is updated
(I could be wrong, but I assume that the tests will for now use the `wmde.5` release...

addshore:
I think this would need to be split into 2 PRs, with the images from the first being used in the second.
Thus I have changed the image in a separate PR now https://github.com/wmde/wikibase-release-pipeline/pull/308
And will now base this PR on that PR

addshore:
Marking as a draft, as https://github.com/wmde/wikibase-release-pipeline/pull/308 would need to be merged first

toban (approved):

PR #308, opened 2022-02-11
Copy LocalSettings.php to shared dir if exists

addshore:


toban (approved):
This seems to work for me when updating. Pushed some docs changes to the followup

PR #313, opened 2022-02-16
wmde.6.release.branch.from.wmde.3

addshore:
- wmde.6 release
- wmde6 Build and test Wikibase and friends


addshore:
Looks like this one then fails as I need to pull out the ENV var changes into this PR too

PR #314, opened 2022-02-16
Fix upgrade

toban:
The env vars from the image being tested was not properly set when
testing leading to the wrong templates being used.

### DIFF
--- a/test/test_upgrade.sh
+++ b/test/test_upgrade.sh
@@ -15,6 +15,7 @@ fi
 
 if [ ! -f "../$TO_VERSION" ]; then
     echo "TO_VERSION does not exist"
+    exit 1
 fi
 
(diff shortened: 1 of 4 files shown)


test/test_upgrade.sh
```diff
@@ -3,14 +3,24 @@
...
 ENV_VERSION=$1
+TO_VERSION=$2
+
+if [ -z "$TO_VERSION" ]; then
+    echo "TO_VERSION is not set"
+    exit 1
+fi
+
+if [ ! -f "../$TO_VERSION" ]; then
+    echo "TO_VERSION does not exist"
```
toban, line 17:
```suggestion
    echo "TO_VERSION does not exist"
    exit 1
```

PR #316, opened 2022-02-17
Split example test into its own workflow

toban:


addshore:
Nice, I was this and was thinking about doing this too!

PR #317, opened 2022-02-18
Github Actions, on PR and push (+selectivity)

addshore:
This makes CI run on PRs as well, rather than only on the
underlying branches.
This will lead to the PR jobs actually running with the base of
the PR merged into the head.

Also be more selective about when to run some jobs, so for example
we can now make docs changes, without running the WHOLE build and
test pipeline.


addshore:
This currently doubles the number of jobs that run, which is probably not ideal.

Perhaps we should only run on `push` on the main branch?
This would lead to PRs always having CI run, and only the main branch having CI run

mariushoch (approved):

addshore:
I still don't understand why this PR makes `Code scanning results / Anchore Container Vulnerability Report (T0)` fail.
I guess I need to do a little more investigation first.

addshore:
The code scanning is just not currently running, (prior to this PR), so perhaps I should just remove it for now...

toban:
> The code scanning is just not currently running, (prior to this PR), so perhaps I should just remove it for now...

It's not running in this PR you mean? I can see it ran on the last commit to main. If you wanna get rid of it completely we probably wanna remove each section of build_and_test.yml it's currently not crashing but makes not sense to keep it.

addshore:
>It's not running in this PR you mean? I can see it ran on the last commit to main.

The last commit on main for me is https://github.com/wmde/wikibase-release-pipeline/actions/runs/1921632522 and I do not see the scan?
At least not in the way it was failing in this PR.

For this PR the scan was running as a totally seperate workflow

image

But indeed, the scan then also normally runs at the end of the other build and test workflows

https://github.com/wmde/wikibase-release-pipeline/blob/main/.github/workflows/built_and_test.yml#L386-L391

So I should probably add the file back.

But I also guess adding the file triggers something else weird (this other workflow we see in this PR) to run sometimes?

addshore:
Think I found it
image

PR #320, opened 2022-02-21
Docs for fed props test systems

addshore:


docs/topics/test-system.md
```diff
@@ -6,16 +6,31 @@ You can find all code for running test test systems in the `/opt/test-systems` d
...
+  - https://wikibase-qs-testing.wmcloud.org (8382 internal)
+
+**Federated Properties**
+
+- Previous release with fed props (internal ports 84**)
+  - https://wikibase-product-testing-fedprops-previous.wmcloud.org (8480 internal)
+  - https://wikibase-query-testing-fedprops-previous.wmcloud.org (8481 internal)
+  - https://wikibase-qs-testing-fedprops-previous.wmcloud.org (8482 internal)
+- Latest release with fed props (internal ports 85**)
+  - https://wikibase-product-testing-fedprops.wmcloud.org (8580 internal)
```
toban, line 31: These seem down which is fine, not sure if it would make sense for anyone merging this to use the script to set it up too rather than the author? (I haven't tried the script out)

toban (approved):
Cool, this looks great

docs/topics/test-system.md
addshore: Heh, these were up yesterday *looks at them*

addshore: Oh wait sorry, yes, I didnt bother setting up this particular test system, as it was not needed for the release that I am currently working through.
If you want to give it a go to setup this instance go for it, but we may run into memory issues and you might need to switch off one of the other systems! (Which I think is fine)

PR #329, opened 2022-03-14
added "Powered by Wikibase" icon in the footer

DavidFichtmueller:
This was discussed in https://phabricator.wikimedia.org/T250009 .
@addshore mentioned:
> Could be worth considering if the wikiba.se website gets its refresh before the next release.

The website has since been refreshed and since I added this piece of code to the LocalSettings my own instance, I thought I might as well create a PR for it.


JeroenDeDauw:
I suspect this is not acceptable due to the image being hosted elsewhere. Local copy would be better.

DavidFichtmueller:
@JeroenDeDauw : Good point. I adjusted the PR. Now the image is stored, when the container is build (in the Dockerfile) and the image paths are local.

addshore:
My only thought here is that if we want this (and I believe we do, but thats up for a product person to decide) then we probably want to just add this to the Wikibase extension!

cc @manicki

addshore:
Closing as per discussions this should be added to the Wikibase extension itself, not this repository.

PR #359, opened 2022-09-01
Stop sharing all mediawiki code between containers

addshore:
This avoids mounting all of the mediawiki & wikibase code into
a volume that will not get updated with upgrades.
Instead option to have the wikibase entry point copy LS.php into
a shared volume, specifically for the docker-compose example if
the directory exists.

This also moves the jobrunner entrypoint into the wikibase image
And refactors the docker-compose files to make use of the same
volumes between the wikibase and jobrunner container easier.

Bug: T298632


docs/topics/upgrading.md
```diff
@@ -107,10 +107,10 @@ Example output
...
 ```sh
 $ docker inspect -f '{{ .Mounts }}' 916ac3ce384e
-[ {volume example_shared /var/lib/docker/volumes/example_shared/_data /var/www/html local rw true } ]
+[ {volume example_shared-localsettings /var/lib/docker/volumes/example_shared-localsettings/_data /var/www/html local rw true } ]
 ```
 
-The above example returns a single mount used by the container called `example_shared`; this is the one that is shared between the jobrunner and the Wikibase web container.
+The above example returns a single mount used by the container called `example_shared-localsettings`; this is the one that is shared between the jobrunner and the Wikibase web container.
 
 We need to remember this name as we will have to remove it manually after the containers has been shut down and removed.
```
addshore, line 115: This should not be the case for the new solution!

addshore:
In the test logs I see lot of `wikibase_jobrunner_1  | LocalSettings.php not shared yet - waiting for 10 seconds.` right now, which explains the ES failure, as the jobs are not running

example/docker-compose.yml
```diff
@@ -49,10 +46,13 @@ services:
       - mysql
     depends_on:
       - mysql
-    restart: always
+      - wikibase
+    restart: unless-stopped
     volumes:
-      - shared:/shared/:ro
       - ./jobrunner-entrypoint.sh:/jobrunner-entrypoint.sh
+      - shared-localsettings:/var/shared-localsettings
```
tarrow, line 53: notably this swap now allow the jobrunner read/write access to the settings file that was previously read only

addshore: I can add :ro back!

addshore, line 53:
```suggestion
      - shared-localsettings:/var/shared-localsettings:ro
```

example/docker-compose.yml
```diff
@@ -25,12 +25,9 @@ services:
     ports:
       - "${WIKIBASE_PORT}:80"
     volumes:
-      ## This shares the configuration with jobrunner
-      - shared:/var/www/html/:rw
-
-      ## Uncomment this next line to mount your own LocalSettings.php file
+      - shared-localsettings:/var/shared-localsettings
+      ## Uncomment this next line to mount your own LocalSettings.php file overrides
       #- ./LocalSettings.php:/var/www/html/LocalSettings.d/LocalSettings.override.php
```
tarrow, line 30: is this legit advice? Won't this cause a deviation between the settings on the job runner and the settings on the web server?

addshore: This comment exists for both the wikibase and job runner containers.
So the expectation would be a user would mount any additional file into both containers

docs/topics/upgrading.md
```diff
@@ -97,41 +97,6 @@ In some newer images, the default value of upload images is written inside the c
 
 Before we do the actual upgrade, we need to stop the containers and remove the volume that is shared between the `wikibase` and `wikibase_jobrunner` containers.
 
-### Review the mounts currently used by the wikibase container
```
tarrow, line 100: I feel like I'm wildly behind the official update process but what makes this no longer necessary? Failing to remove this volume would presumably prevent "recreation" of the LocalSettings?

addshore: > Failing to remove this volume would presumably prevent "recreation" of the LocalSettings

Why do we want LocalSettings.php to be recreated?
We do not want install.php to run a second time, so LocalSettings.php needs to remain?

docs/topics/upgrading.md
```diff
@@ -97,41 +97,6 @@ In some newer images, the default value of upload images is written inside the c
 
 Before we do the actual upgrade, we need to stop the containers and remove the volume that is shared between the `wikibase` and `wikibase_jobrunner` containers.
```
tarrow, line 98: Looks like this should go too

addshore: Gone!

example/docker-compose.yml
addshore: Altered in https://github.com/wmde/wikibase-release-pipeline/pull/359/commits/75631dd9db9e8782979352d3340af0742fdb1b10 to only appear in 1 place

rti:
Problem fixed in new WBS Deploy, this should not be relevant anymore. In case it is, feel free to reopen

PR #366, opened 2022-09-12
publishing.md docs, load variables.env before release env

manicki:
Wrong order of source commands might generated not wanted package names that get published


manicki:
@addshore suggests this might be not wanted change. Publishing as a documentation

addshore:
From mattermost

```
@Leszek Manicki infact, 100% they are the wrong way around

infact, its broken the way around it currently is i believe

because https://github.com/wmde/wikibase-release-pipeline/blob/main/versions/wmde7.env#L26 would get overridden by variable.env which would be undesired too
```

PR #368, opened 2022-09-13
wmde.7 docker-compose example

addshore:


addshore:
I think we only link to a latest docker-compose so let me check that

addshore:
Will open a new PR from a new branch...

PR #376, opened 2022-10-31
Fixed the RELEASE_MAJOR_VERSION in versions/wmde9.env

manicki:
Causes publishing tarball packages to a wrong directory


manicki:
I wish I understood why one CI run times out constantly.

PR #390, opened 2023-01-02
Add localsettings comment to wikibase_jobrunner

schelv:
Having the jobs run with the same settings as wikibase is probably what wou want.


addshore:
A comment like this probably makes sense for the code in main right now.

PR #402, opened 2023-01-23
Add debug_extra_env var for triggerable builds

addshore:


tarrow (approved):
looks fine to me;

rather shotgun surgery suggest we ought to DRY out this file if we can at some point though

addshore:
> rather shotgun surgery suggest we ought to DRY out this file if we can at some point though

Yes, this file is giant and confusing

PR #408, opened 2023-03-15
wmde11: Set error_reporting in wikibase php.ini file

addshore:
- Ignore deprecation warnings in wikibase php.ini
- Remove error_reporting setting from localsettings files
- Note in changes.md


manicki (approved):
I figure it does not matter practically at all in reality but given the deprecation notices come from wikibase/mediawiki, woudln't it be "cleaner" to have the error reporting setting tweaked in medawiki config rather than in overall PHP config?

(no point it changing it, as it is a temporary situation any way)

addshore:
Setting this in the MediaWiki config may be nicer.
However many users will override the config, or the template, so that they can configure their wikibase.
As a result none of these users would "get" the change, and would thus see deprecation warnings on upgrade.
The nicer fix here would be to backport the changes that avoid the deprecation warnings in php 8 in the first place.
We considered this earlier in the release process and decided not to due to the number of backports.

PR #419, opened 2023-06-05
T341472: use /usr/bin/env in shebang

rti:
Until now, scripts hardcoded the path to bash in the shebang line, rendering the scripts unable to execute on systems where bash does not live in /bin/bash (e.g. NixOS). This change introduces the use of env(1) in shebangs in order to lookup bash in the current environment.


lorenjohnson (approved):
✅

adeeri (approved):

PR #420, opened 2023-06-05
T341653: use docker compose v2

rti:
docker-compose V1 is replaced by docker compose V2, V1 is deprecated.
https://docs.docker.com/compose/#compose-v2-and-the-new-docker-compose-command


rti:
Apparently, autogenerated container names change in between docker compose v1 and v2. This appears to be the reason this PR fails in CI now. #431 contains some fixes for that which might be ported back to this PR at some until then this PR must not be merged.

lorenjohnson (approved):
🪼

adeeri (approved):

PR #421, opened 2023-06-05
T341222: extend gzip compression rate usage

rti:
The release pipeline has a setting called GZIP_COMPRESSION_RATE that can be set in local.env. Using this setting gzip invocations are configured. The user can choose the compression rate/speed tradeoff. 1 being fast and weak compression, 9 being slow and strong compression. This change extends the usage of this configuration value to the remaining gzip invocations as well as fixes the handing over of the actual value to gzip by removing the extra 'f' (not required or supported according to man gzip(1)).

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,4 +13,3 @@ For [testing](docs/topics/testing.md), you can use `test`, `test-all` make targe
 ## Documentation
 
 The pipeline documentation can be found [here](docs/index.md).
-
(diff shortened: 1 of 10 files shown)


lorenjohnson (approved):
Weird about the "f", almost as if someone was trying to cat the value to Float? Anyway, good stuff. LGTM

adeeri (approved):

PR #423, opened 2023-06-15
fix: test failures

rti:
Just some experimental test failure fixes / work arounds.


rti:
@lorenjohnson just mentioned:

`docker ps -q | xargs docker stop; docker container prune --force; docker volume ls -q | xargs docker volume rm;` on test start would do most of the tricks as well.

rti:
superseeded by #431

PR #424, opened 2023-06-19
Add an Elasticsearch image with a Bitnami base

AndrewKostka:


addshore:
As an outsider is rather hard to see why this is needed or what it relates to?
Are there any phab tasks / description for this etc?

PR #428, opened 2023-07-03
T341654: git/docker ignore mwlog folder

rti:
It should not be commited nor put into image builds.


lorenjohnson (approved):
Ah yes, glad to see this one 👌🏼

adeeri (approved):

PR #434, opened 2023-07-11
Update extension list in WikibaseBundle README.md

manicki:


rti (approved):
thanks a lot!

manicki:
@rti note that CirrusSearch + WikibaseCirrusSearch extensions have actually been listed already, in the same line as the Elastica extension. Questionable way of doing it but I believe the motivation was that they three only make sense used together in Wikibase context

rti:
> @rti note that CirrusSearch + WikibaseCirrusSearch extensions have actually been listed already, in the same line as the Elastica extension. Questionable way of doing it but I believe the motivation was that they three only make sense used together in Wikibase context

good catch, reverting my commit.

adeeri (approved):

PR #431, opened 2023-07-04
T340227: Run tests consistently

rti:
https://phabricator.wikimedia.org/T340227

PR #435, opened 2023-07-14
T340939: wmde14 based on mediawiki 1.40.0

rti:
https://phabricator.wikimedia.org/T340939


rti:
based on #431 for more consistent pipeline behaviour, review #431 first!

GreenReaper:
Regarding [this commit](https://github.com/wmde/wikibase-release-pipeline/pull/435/commits/b905a8d1a5e41171cfd2220f1ebcc9d1bfb046e5) - the 'f' is not part of the compression level value but a separate option, i.e. `gzip -9f` is equivalent to `gzip -9 -f`, so it wasn't preventing the level from being set.

I agree that it is probably not _useful_ in this case, though, because you are operating on files through piping redirection (I believe gzip uses isatty() to detect this) and so it won't need to ask about overwriting files or dealing with too many levels of symbolic links.

rti:
superseeded by #478

PR #442, opened 2023-08-03
Wait-for-jobs

lorenjohnson:


Docker/test/curl/test_curl.sh
```diff
@@ -5,17 +5,24 @@ check_if_up() {
...
+    if ! (/usr/bin/curl --fail --output /dev/null --silent "$FULL_URL";) then
+        echo "🔄 Waiting for $DISPLAY_URL"
+
+        if ! (/usr/bin/curl --fail --retry 120 --retry-all-errors --retry-delay 1 --max-time 10 --retry-max-time 120 --output /dev/null --silent "$FULL_URL";) then
+            echo "❌ Could not load $DISPLAY_URL"
+            exit 1
+        fi
     fi
+
+    echo "✅ Successfully loaded $DISPLAY_URL!"
```
rti, line 25: Found it a bit distracting to have the ! at the end of the url, works without as well, I guess. Love the emojis btw :)

```suggestion
    echo "✅ Successfully loaded $DISPLAY_URL"
```

Docker/test/curl/test_curl.sh
```diff
@@ -5,17 +5,24 @@ check_if_up() {
     PATH=$2
 
     FULL_URL="$1$2"
+    DISPLAY_URL="http://$FULL_URL"
 
     if [ -z "$URL" ]; then
+        echo "❌ check_if_up called with an empty URL (with PATH: $2)"
```
rti, line 11:
```suggestion
        echo "❌ check_if_up called with an empty URL (with PATH: $PATH)"
```

makes it a bit more readable

Docker/test/curl/test_curl.sh
```diff
@@ -26,6 +33,4 @@ if [ -z "$SKIP_WDQS" ]; then
 
     check_if_up "$WDQS_FRONTEND_SERVER" "/"
 
-fi
-
-
+fi
```
rti, line 36: Could we agree on having a `\n` at the end of every file? Does vscode support that? Otherwise I think we will have lots of those whitespace changes back and forth in the future.

POSIX wants that too https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206

Docker/test/selenium/helpers/default-functions.js
```diff
@@ -223,6 +223,35 @@ const defaultFunctions = function () {
...
 
+	browser.addCommand( 'waitForJobs', async function ( {
+		serverURL = process.env.MW_SERVER,
+		timeout: providedTimeout,
+		timeoutMsg
+	} = {} ) {
+		let jobsInQueue;
+		// timeout is 1 second less than default Mocha test timeout
+		const timeout = providedTimeout ||
+			( process.env.MOCHA_OPTS_TIMEOUT ? process.env.MOCHA_OPTS_TIMEOUT - 1000 : 89000 );
```
rti, line 234: Should we document the 89000 magic number here as well?

test/docker-compose.yml
```diff
@@ -1,6 +1,19 @@
 # Wikibase for testing
 version: '3'
 
+x-common-variables: &wikibase_variables
```
rti, line 4: I wonder whether moving all variables including passwords (that then end up in all containers) is a good pattern. I see the point of improved clarity in the files though.

test/docker-compose.yml
```diff
@@ -20,17 +34,29 @@ services:
...
+    command:  /jobrunner-entrypoint.sh
+    links:
+      - mysql
+    depends_on:
+      - mysql
+    restart: always
+    volumes:
+      - shared:/shared/:ro
+      - ../example/jobrunner-entrypoint.sh:/jobrunner-entrypoint.sh
+      ## Uncomment this next line to mount your own LocalSettings.php file
```
rti, line 50: I think this comment is misleading in the test docker compose yml

rti (changes requested):
Oh, this is beautiful. Love to see all those waits being thrown out.

test/docker-compose.yml
lorenjohnson: Had kept to keep it looking like example, but generally agree and will remove the comments here for now.

Docker/test/curl/test_curl.sh
lorenjohnson: Agreed. I thnk the "!" was just carried over from the original implementation.

Docker/test/curl/test_curl.sh
lorenjohnson: I would prefer a required linebreak at the end of every file. I don't have my settings in VS Code on this project do that and was actually removing them when I found them as it seemed the default in this project was to not have a trailing \n. Doesn't seem to be any linting about it, and like I say, I prefer (is more familiar to me) to have a final \n. Probably won't change anything here now though as we sort this out more generally.

rti: follow up in https://phabricator.wikimedia.org/T343802

Docker/test/selenium/helpers/default-functions.js
rti: beautiful

test/docker-compose.yml
rti: Yes, still questioning whether this is a good idea in general, especially for "the distribution", probably ok in this case though.

adeeri (approved):

test/docker-compose.yml
lorenjohnson: For sure. These "secrets" should probably be stored in a different way and I think there is a Docker way to do this. This item should be highlighted on the "explore wikibase deployment to production" spike. In the case of this container though it does needs all these variables and they're the same set we were passing a slightly different way before. Nothing was added or changed here in terms of the existing security implications.

lorenjohnson:
@rti I think after I made changes as per your notes I requested a "re-review". Do you mind re-reviewing and/or approving now? :) I think we may be ready to merge this one.

rti (approved):

PR #443, opened 2023-08-05
T343347 (2): Quiet STDOUT test output, Improve test log and screenshot file organization

lorenjohnson:
https://phabricator.wikimedia.org/T343347

- Consolidates all test logs and Selenium screenshots to `test/log/$SUITE` (the upgrade test creates 3 directories, pre_upgrade, upgrade, and post_upgrade)
- Deletes logs for the related suite on each test run
- `make test` (without a `SUITE=` specified) now defaults to running `make test-all`
- Eliminates "WikibaseApi not initialised warnings" by initialising once in Mocha global `before` for all test runs
-  Hides WDIO-Sync warning by changing `SELENIUM_LOG_LEVEL` to default to `error` instead of `warn`, if you want to see WDIO warnings you can re-enable locally by setting `SELENIUM_LOG_LEVEL=warn`
- Removes `wikibase-selenium-test`, `curl`(now `test-setup`), and `builder` containers after use with a `run --rm` to keep dead Docker containers from piling-up

**TEST OUTPUT SAMPLE**

Screenshot 2023-08-08 at 10 34 10

**LOG FILE ORGANIZATION LAYOUT SAMPLE**

Screenshot 2023-08-07 at 23 06 01


test/docker-compose-selenium-test.yml
```diff
@@ -9,7 +9,7 @@ services:
     depends_on:
       - wikibase
     volumes:
-      - ./log/selenium/:/usr/src/app/log/
+      - ../test/log:/usr/src/app/log
```
lorenjohnson, line 12: This is a bit of a functional hack for the case of `test-example` but still works same as before in all other cases.

test/reporter/report.js
```diff
@@ -7,7 +7,7 @@ if( !process.env.SUITE ) {
     return;
 }
 
-const filePath = '../log/selenium/result-'+ process.env.SUITE + '.json';
+const filePath = `../log/${process.env.SUITE}/selenium-result.json`;
```
lorenjohnson, line 10: This path actually gets overridden by the WDIO JSON reporter to `log/$SUITE/result.json`. This was the case before they branch, and this may be best removed here or otherwise commented for clarity.

.github/workflows/built_and_test.yml
```diff
@@ -544,8 +544,8 @@ jobs:
         if: always()
         continue-on-error: true
         run: |
-          ls -lahr test/mwlog test/mwlog/*
-          tail -n +1 test/mwlog/*/*.log
+          ls -lahr test/log test/log/*
+          tail -n +1 test/log/*/*.log
```
adeeri, line 548: Might be a good chance to change this file's name :)

Makefile
```diff
@@ -10,16 +10,20 @@ download:
 	bash publish/download.sh
 
 .PHONY: test
-test: test-stop
-	bash test/test_suite.sh ${SUITE}
+test:
```
adeeri, line 13: Doesn't it still need to depend on `test-stop`?

lorenjohnson: It doesn't anymore because of these changes @roti made awhile back which stops, and removes the `test` containers and volumes before each test run: https://github.com/wmde/wikibase-release-pipeline/blob/244f2e363ab7d8f27f1a64eb4e591bde3e92d701/test/run_selenium.sh#L34C15-L36

`test_stop.sh` can and probably should be removed for clarity though.

.github/workflows/built_and_test.yml
lorenjohnson: Not sure what you mean in terms of the file's name? 

_However line 548 here maybe should be changed to `tail` only `test/log/*/wikibase/*.log` to have the same behaviour as before. Wherever this log tailing appears this currently will also be tailing the logs Docker log and setup log... Not necessarily wrong, but that is a lot of logging and different than before. Hmm... I am ok leaving this, but_

adeeri: I meant `built` -> `build`.

As for logging, if it's not too much clutter, I don't see any harm in also having the non-wikibase logs.

rti: Are you still planning to rename the file? The conversation was already resolved, but as I see it states still "built" in the filename.

Makefile
rti: Should we remove `test-stop.sh` and the make targets in this PR then?

test/test_suite.sh
```diff
@@ -29,17 +31,23 @@ else
     WIKIBASE_TEST_IMAGE_NAME="$WIKIBASE_BUNDLE_IMAGE_NAME"
 
     # load additional bundle images
-    docker load -i "../artifacts/$ELASTICSEARCH_IMAGE_NAME.docker.tar.gz"
-    docker load -i "../artifacts/$QUICKSTATEMENTS_IMAGE_NAME.docker.tar.gz"
+    {
+        docker load -i "../artifacts/$ELASTICSEARCH_IMAGE_NAME.docker.tar.gz" >> "$SETUP_LOG" 2>&1
+        docker load -i "../artifacts/$QUICKSTATEMENTS_IMAGE_NAME.docker.tar.gz" >> "$SETUP_LOG" 2>&1
+    } >> "$SETUP_LOG" 2>&1
```
rti, line 37: I think either the individual or the joined `>> "$SETUP_LOG" 2>&1` can be removed

test/test_suite.sh
```diff
@@ -3,21 +3,23 @@ set -e
...
+
 if [ -z "$SUITE" ]; then
-    echo "SUITE is not set"
+    echo "🚨 SUITE is not set"
     exit 1
 fi
 
+# setup log directory, create "last-ran" file and setup log
+rm -Rf "log/$SUITE"
+mkdir -p "log/$SUITE/wikibase"
```
rti, line 19:
```suggestion
mkdir -p "log/$SUITE/wikibase"
mkdir -p "log/$SUITE/client"
```
... should fix running the `repo_client` suite

test/test_example.sh
```diff
@@ -19,4 +19,10 @@ export DATABASE_IMAGE_NAME="$MYSQL_IMAGE_NAME"
 ## Use in combination with example compose files 
 export DEFAULT_SUITE_CONFIG="-f ../example/docker-compose.yml -f ../example/docker-compose.extra.yml -f docker-compose.example.yml"
 
+# setup log directory, create "last-ran" file and setup log
+rm -Rf "log/$SUITE"
+mkdir -p "log/$SUITE"
```
rti, line 24:
```suggestion
mkdir -p "log/$SUITE/wikibase"
mkdir -p "log/$SUITE/client"
```

should fix the permission denied thingy

rti (changes requested):

test/test_suite.sh
lorenjohnson: Interestingly I added the grouping as our linting wouldn't allow two lines to have the same output without this. Can't remember the logic as to why, but was fine with it :)

Makefile
lorenjohnson: I was being cautious, but have went through it on this and will remove in this branch as well.

test/test_example.sh
lorenjohnson: This I think does fix it on the Thinkpads, but running as the current user breaks things on Github Actions CI. I've made additional changes in a subsequent branch which moves the screenshot and result file manage directly into the wikibase-selenium-test container such that the same user (root) that creates the files also deletes them, which logically is a better bet. I have not yet done the same with the MediaWiki/Wikibase containers for those logs, but those for whatever reason don't seem to be causing any issue. The other logs are created and maintained by the shell scripts (`log/SUITE/setup.log` and `log/SUITE/SUITE.log`). 

I've created a follow-up PR for this additional changes here: https://github.com/wmde/wikibase-release-pipeline/pull/445

lorenjohnson:
Closing in preference to: https://github.com/wmde/wikibase-release-pipeline/pull/445

PR #445, opened 2023-08-13
T343347: Quieter and more clear test output

lorenjohnson:
https://phabricator.wikimedia.org/T343347

⚠️ Before testing this branch locally I recommend doing the following:

1) Delete all old test files directories:

`sudo rm -Rf test/log && sudo rm -Rf test/mwlog`

2) Remove any currently running Docker containers and volumes:

`docker ps -q | xargs docker stop; docker container prune --force; docker volume ls -q | xargs docker volume rm;`

----

* Makes on-screen test run logging more concise, clear, and consistently focused on test results (previous Docker operations detail is available at the top of `test/log/SUITE/SUITE.log`)
* Updates the WDIO Spec reporter and removes the `[0-0] RUNNING in chrome...` log prefix to the WDIO Spec reporter (the final test report that prints out)
* Hides WDIO-Sync warning by changing `SELENIUM_LOG_LEVEL` to default to `error` instead of `warn`, if you want to see WDIO warnings you can re-enable locally by setting `SELENIUM_LOG_LEVEL=warn`
* Eliminates "WikibaseApi not initialised warnings" by initialising once in Mocha global `before` for all test runs
* Consolidates all test logs and Selenium screenshots to `test/log/$SUITE` (the upgrade test creates 3 directories, pre_upgrade, upgrade, and post_upgrade)
* `make test` (without a `SUITE=` specified) now defaults to running `make test-all`
* Deletes logs for the related suite on each test run
* Renames Github Action test runs to `test_wikibase (repo, mariadb...)` from `test_wikibase (mariadb..., repo)`, making it easy to see the name of the test suite ran in the summary results here on Github.
* Refactors test run shells scripts in `test/*.sh` to be somewhat more clear and consistent with commenting
* Changes WDIO logging to default to `error` vs `warn` to reduce noise. Set `SELENIUM_LOG_LEVEL=warn` in `local.env` if you want the warnings back (i.e. "`wdio/sync` deprecation warnings)
* Sets a default `MAX_JOBS` to 2 `test/default.env` to prevent jobrunner long execution time, and silence warnings when its not set
* Renames `Docker/test/curl` to `Docker/test/setup` to better reflect its purpose. 
* Fixes an issue such that job runner wasn't configured correctly for the elasticsearch test suite, but was sometimes passing anyway because "jobs on request" was still on in the test environment (which this branch turns off)
* Increases `check_if_up` `curl` timeout to to ~240 seconds from ~120 seconds, which often wasn't enough for slower Github Action job runners to get Wikibase up.
- Removes `wikibase-selenium-test`, `curl`(now `test-setup`), and `builder` containers after use with a `run --rm` to keep dead Docker containers from piling-up

Also relates to understanding the upgrade tests better: https://phabricator.wikimedia.org/T343870#9088696


test/docker-compose-selenium-test.yml
```diff
@@ -6,8 +6,6 @@ services:
     build:
       context: ../Docker/test/selenium/
       dockerfile: Dockerfile
-    depends_on:
-      - wikibase
```
lorenjohnson, line 10: Note: This was extraneous and perhaps not strictly correct, it makes it impossible to docker compose with this file for taking the test runner service down when the other services aren't running. It had no impact to remove it in terms of the test runs.

.nvmrc
```diff
@@ -0,0 +1 @@
+14.21.3
```
lorenjohnson, line 1: I would recommend we all use NVM (Node Version Manager) with a shell addition to automatically switch to the project version of Node when switching into any directory in the project. Either way this file makes that possible and is how I run in my dev environment: https://github.com/nvm-sh/nvm

Docker/test/selenium/wdio.conf.js
```diff
@@ -49,19 +57,33 @@ exports.config = {
...
-	// Setting this enables automatic screenshots for when a browser command fails
-	// It is also used by afterTest for capturig failed assertions.
-	screenshotPath: process.env.LOG_DIR || __dirname + '/log',
+	// Setting this enables automatic screenshots for when a browser command fails assertions.
+	screenshotPath,
 
 	// Default timeout for each waitFor* command.
 	waitforTimeout: 30 * 1000,
 
 	// See also: http://webdriver.io/guide/testrunner/reporters.html
```
adeeri, line 68: This seems broken and also unnecessary? Not related to your changes, but maybe a chance to delete/update.

lorenjohnson: Will catch in the code organisation branch.

.github/workflows/build_and_test.yml
```diff
@@ -507,6 +506,7 @@ jobs:
           base__pingback,
           base__fedprops,
         ]
+        databaseImageName: [ 'mariadb:10.9' ] # 'mysql:5.6' disabled https://phabricator.wikimedia.org/T296066
```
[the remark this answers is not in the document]
lorenjohnson: Yes, this is what changes the order in the Gihub Actions run from (mariadb-...., SUITE) to (SUITE, mariadb....)

Docker/test/selenium/wdio.conf.js
```diff
@@ -49,19 +57,33 @@ exports.config = {
 	// ===================
 
 	// Level of verbosity: "trace", "debug", "info", "warn", "error", "silent"
-	logLevel: process.env.SELENIUM_LOG_LEVEL || 'warn',
+	logLevel: process.env.SELENIUM_LOG_LEVEL || 'error',
```
RickiJay-WMDE, line 60: Do we want a --verbose option to go back to 'warn', if not output everything that is now in the logs?

lorenjohnson: That would be nice, but don't want to add the command line arg parsing between the Makefile and Shell Scripts for that (and a couple other options) until things settle a bit more. It would be slightly more than trivial to add that right now, especially given the cost of "testing test runs". All for this being a feature ticket for future :)

.github/workflows/build_and_test.yml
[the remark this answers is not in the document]
adeeri: Cool, just meant the functionality here is the same

adeeri (approved):

RickiJay-WMDE (approved):

PR #452, opened 2023-08-19
T344551: Don't load Elasticsearch extensions when Elasticsearch isn't present

lorenjohnson:
https://phabricator.wikimedia.org/T344551

### DIFF
--- a/Docker/build/WikibaseBundle/LocalSettings.d.template/CirrusSearch.php
+++ b/Docker/build/WikibaseBundle/LocalSettings.d.template/CirrusSearch.php
@@ -1,6 +1,8 @@
 <?php
 
-// https://www.mediawiki.org/wiki/Extension:CirrusSearch
-wfLoadExtension( 'CirrusSearch' );
+if ( getenv('MW_ELASTIC_HOST') !== false ) {
+  // https://www.mediawiki.org/wiki/Extension:CirrusSearch
+  wfLoadExtension( 'CirrusSearch' );
 
-// See WikibaseCirrusSearch.php for further configuration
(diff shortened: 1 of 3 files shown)


adeeri (commented):
I don't know enough to be sure that these extensions are not needed for some reason even when elastic search isn't set, but basically lgtm considering the log clutter

lorenjohnson:
> I don't know enough to be sure that these extensions are not needed for some reason even when elastic search isn't set, but basically lgtm considering the log clutter

At this point, I do think I know enough to say that all 3 of these extensions are only ever used for Elasticsearch purposes and don't have any utility (and in fact break things) if ran without Elasticsearch available. I would be ok moving ahead with merging this, and am also very ok running it past others who've been around longer outside our team for a double check.

lorenjohnson:
@tarrow do you know of any reason not not have the `CirrusSearch`, `Elastica`, and `WikibaseCirrusSearch` MediaWiki extensions disabled when Elasticsearch isn't running? It seems clear enough to me that they shouldn't be added in this case, which I assume was just an oversight, but would like a confirmation that I'm not missing something. As-is this causes an excess of Elasticsearch write failures in the MediaWiki logs even though the default MySQL search does still work.

manicki:
i am not @tarrow but conceptually it does occur to be an oversight 100%.
If there is a "technical" reason why Wikibase would need said extensions to be present to provide non-search functionality that'd indicate some sloppy engineering around dependency control in Wikibase code IMO, that deserves to be fixed.

adeeri (approved):

PR #457, opened 2023-08-25
T344696 Pre-Commit Newline Hook

RickiJay-WMDE:
As is: this only runs on staged files of specified extensions. Will update such files with newlines. Per discussion in standup 08.28.2023, will _not_ add these changes to the commit, as doing so would add all unstaged changes.

Please squash this merge - most of these commits are just trying something to see if it passed lint.


test/.husky/pre-commit
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env sh
+# shellcheck disable=SC1091
+. "$(dirname -- "$0")/_/husky.sh"
```
RickiJay-WMDE, line 3: To explain: running the shell lint checker on this threw a SC1091, file not found, error on `/_/husky.sh`. The alternative to this would be installing husky on the pipeline, but that seemed excessive and unnecessary.

PR #458, opened 2023-08-28
T343802 Fix: Run linting only in Docker container, and quiet status messages

lorenjohnson:
Follow-up to original T343802 work. 
* Runs all linting through Docker
* Turns-off the extraneous "ℹ️  Linting ..." messages, it will only output anything to STDOUT if there are linting errors
* Moves `lint.sh` from root of project into `test/scripts`

It may be worth iterating this further with a git pre-commit in mind. That would probably look like adding `hadolint` and `spellcheck` to our test-runner (`test/Dockerfile`), and only running `lint.sh` directly from that container (`docker compose run --rm test-runner -c "scripts/lint.sh`). `husky` is the issue, as it's an npm package and I'm not sure we can bypass that (this strongly relates to work in https://github.com/wmde/wikibase-release-pipeline/pull/457)


test/Dockerfile
```diff
@@ -1,10 +1,15 @@
 FROM node:14.21.3-bullseye
 
 RUN apt-get update && \
-    apt-get -y install docker.io chromium python3-pip && \
+    apt-get --no-install-recommends -y install \
+        docker.io=20.10.5+dfsg1-1+deb11u2 \
```
lorenjohnson, line 5: The extra version specifications here and in the `pip3` install below are here to quiet warnings. Regardless I do think for this test-runner container it's best if we pin versions anyway.

test/scripts/lint.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
lorenjohnson, line 1: Sorry. Surprised Git didn't pickup this file move. I guess I made too many changes at the same time as the move, so it lost track 🤷🏼‍♂️

RickiJay-WMDE (approved):

test/Dockerfile
```diff
@@ -2,14 +2,14 @@ FROM node:14.21.3-bullseye
...
+        docker.io \
+        chromium \
+        python3-pip && \
     ln -sf /usr/bin/python3 /usr/bin/python && \
     rm -rf /var/lib/apt/lists/* && \
     pip3 install --no-cache-dir --upgrade \
-        pip==23.2.1 \
-        setuptools==68.1.2
+        pip \
+        setuptools
```
RickiJay-WMDE, line 12: Do we not need to pin versions after all, then?

lorenjohnson: Nope, and you were right, we probably shouldn't! 

There were a series of small issues here starting with the path to the `.hadolint` config file being wrong, hence causing linting warnings which otherwise the `.hadolint` config had turned off that led to me following and pinning the versions 🤦🏼‍♂️ It's all sorted now, and hopefully simpler to read. Gonna request a re-review, if you don't mind.

rti (approved):
LGTM 🌻

RickiJay-WMDE (approved):

PR #460, opened 2023-08-30
T343348 Fix: Test run status logging pointing to wrong path to test log file

lorenjohnson:
A very minor issue with the log file path in the case of a `test/suites/SUITE/setup.sh` existing. The status message that logs that it is using that file was not writing to log as the path was wrong. This fixes that.

### DIFF
--- a/test/suites/setup.sh
+++ b/test/suites/setup.sh
@@ -8,6 +8,6 @@ scripts/check_if_up.sh "$WDQS_FRONTEND_SERVER" "/"
 SUITE_SETUP_FILE="suites/$SUITE/setup.sh"
 
 if [ -f "$SUITE_SETUP_FILE" ]; then
-    echo "🔄 Running \"$SUITE_SETUP_FILE\"" 2>&1 | tee -a "$TEST_LOG"
+    echo "🔄 Running \"$SUITE_SETUP_FILE\"" 2>&1 | tee -a "../$TEST_LOG"
     $SUITE_SETUP_FILE
 fi


test/suites/setup.sh
```diff
@@ -8,6 +8,6 @@ scripts/check_if_up.sh "$WDQS_FRONTEND_SERVER" "/"
 SUITE_SETUP_FILE="suites/$SUITE/setup.sh"
 
 if [ -f "$SUITE_SETUP_FILE" ]; then
-    echo "🔄 Running \"$SUITE_SETUP_FILE\"" 2>&1 | tee -a "$TEST_LOG"
+    echo "🔄 Running \"$SUITE_SETUP_FILE\"" 2>&1 | tee -a "../$TEST_LOG"
```
rti, line 11: This is running in the context of the `test-runner`, so the `$TEST_LOG` var is not available and the path to the logfile is also not valid.

I think the solution to properly fill this into the logging mechanism is quite simple. As [run_selenium.sh](https://github.com/wmde/wikibase-release-pipeline/blob/abc4064e1fb1c901b4f79247819df229aaf63220/test/scripts/run_selenium.sh#L33) calls this script already with stderr-to-stdout redirection and pipes it through `tee`, we can completly remove the redirection thingy here. It is done one layer above already.  

```suggestion
    echo "🔄 Running \"$SUITE_SETUP_FILE\""
```

rti (changes requested):

test/suites/setup.sh
lorenjohnson: @rti ok, so I looped back and see the code now for my solution. I want to keep the *.sh scripts using the same patterns throughout to make them as easy as possible to reason about. Also I wasn't previously passing in `RESULTS_DIR` either which `wdio.conf.js` needed. That wasn't causing any trouble as the default value when that wasn't present was duplicating the `RESULTS_DIR` composition. I've removed that default to keep this all happening in one place (the runner *.sh scripts). Hope that makes sense. Your solution was fine too, but this other bit of housekeeping made a bit more sense to me in the overall setup 🤓

rti (approved):
works for now. thanks a lot for the update.

as soon as more test setup and tear down logic moves into wdio, we might get rid of some of this convoluted shell scripts anyway.

PR #461, opened 2023-08-31
T344696 Pre-Commit Lint

RickiJay-WMDE:
Run the linting script before each commit
Add a specific check for newlines at the end of files


lorenjohnson (commented):
I think it would be good to make all the newline additions as part of this PR if that isn't what you're already doing. Surprised that only these few files came-up new-line'less. 

Otherwise looks great and will approve after I get that answer :)

RickiJay-WMDE:
@lorenjohnson This appears to be everything within /test, but if you know of any files I'm missing, please point me in their direction?

lorenjohnson:
> @lorenjohnson This appears to be everything within /test, but if you know of any files I'm missing, please point me in their direction?

Ah yes, I think the expectation is that it checks for newlines on those file types across the whole repo. You may want to make where you ignore `node_modules` to check against an array of ignored directories, or otherwise utilise `.gitignore` in each directory recursively. Happy to pair/talk this out or Robert I think will be able to help.

RickiJay-WMDE:
Found another 67 files, @lorenjohnson , and I added a Python formatter to the test runner
@rti  and I put our heads together for the EXTENSIONS and EXCLUSIONS lists in `add_newline.py`, please let me know if I missed anything huge

rti (changes requested):

test/scripts/add_newline.py
```diff
@@ -0,0 +1,61 @@
...
+
+def is_excluded(path: str) -> bool:
+    for exclusion in EXCLUSIONS:
+        if exclusion in path:
+            return True
+    return False
+
+
+def file_should_be_run(file: str) -> bool:
+    if file == "Dockerfile":
```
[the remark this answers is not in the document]
RickiJay-WMDE: Switched to file regex

test/scripts/lint.sh
```diff
@@ -16,3 +16,10 @@ docker compose -f test/docker-compose.yml run --rm -v "$(pwd)/docs/diagrams:/tmp
   npm install --loglevel=error --progress=false --no-audit --no-fund > /dev/null &&
   npm run lint --silent
 "
+
+# ℹ️ Linting newlines across the repo, and formatting Python scripts
+docker compose -f test/docker-compose.yml run --rm -v "$(pwd):/tmp" test-runner -c "
+  pip3 install black &&
+  python3 -m black /tmp/ &&
```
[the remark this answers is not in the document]
RickiJay-WMDE: Removed `black`

test/scripts/add_newline.py
```diff
@@ -0,0 +1,59 @@
...
+
+        contents = current_file.read(1)
+        if not contents.endswith("\n"):
+            current_file.write("\n")
+            return True
+    return False
+
+
+if __name__ == "__main__":
+    for root, _, files in os.walk("/tmp"):
```
rti, line 55: please make the path a parameter to the script

test/scripts/add_newline.py
```diff
@@ -0,0 +1,59 @@
...
+
+def add_newline(file: str, root_dir: str) -> bool:
+    file_path = os.path.join(root_dir, file)
+
+    with open(file_path, mode="r+") as current_file:
+        current_file.seek(0, os.SEEK_END)
+        current_file.seek(current_file.tell() - 1, os.SEEK_SET)
+
+        contents = current_file.read(1)
+        if not contents.endswith("\n"):
```
rti, line 48: checking for equality here should be enough

test/scripts/add_newline.py
```diff
@@ -0,0 +1,59 @@
...
+    ".+\.lua",
+    ".+\.sql",
+    ".+\.sx",
+    ".+\.php",
+    ".+\.xml",
+    ".+\.md",
+    ".+\.template",
+    "Dockerfile",
+]
+EXCLUDE_PATHS = ["node_modules", ".git", "git_cache", "cache", "artifacts"]
```
rti, line 23: maybe we should add a note that this only works for directories here? or maybe just rename the var to `EXCLUDE_DIRS`?

rti (changes requested):

test/scripts/add_newline.py
RickiJay-WMDE: Renamed to EXCLUDE_DIRS

test/scripts/add_newline.py
RickiJay-WMDE: Changed to equality

rti (approved):
LGTM. Thanks for your work on that!

RickiJay-WMDE:
> Can't we handle the exclusions through inheriting (what I believe is) the same logic as our `.gitignore`s? I think that getting the git index for the entire repo and passing that into the newlines check is best, and achievable ? 

So the problem was that git doesn't exist inside the docker container. If I try `print(os.popen('git status'))`, I get `fatal: not a git repository`.

The solution ended up being using a shell variable for the file list that I could then pass to Python, which I had no idea how to do until now, so thanks for making me learn that.

lorenjohnson:
> > Can't we handle the exclusions through inheriting (what I believe is) the same logic as our `.gitignore`s? I think that getting the git index for the entire repo and passing that into the newlines check is best, and achievable ?
> > @lorenjohnson
> 
> So the problem was that git doesn't exist inside the docker container. If I try `print(os.popen('git status'))`, I get `fatal: not a git repository`.
> 
> The solution ended up being using a shell variable for the file list that I could then pass to Python, which I had no idea how to do until now, so thanks for making me learn that.

I gotcha, but there are a couple ways through that which I think aren't bad: 1) It would be fine with me though if we added git to that container, or, 2) we could pass this as an ENV VAR (`GIT_LINT_FILES` or some such) to the container by assigning it in `precommit` where we still have git and then adding that to the environment variables with passthrough to test runner in `test/docker-compose.yml`.  Certainly prefer the first option, and think it's worth doing. IMHO as always :)

lorenjohnson:
> 

For sure, just chatting. What is here is fine, and if you are up for making this improvement and want a hand I'm happy to help with the Docker steps or otherwise get into with you.

lorenjohnson (approved):
Ah--hadn't realized you had already made the change. Looks great!

PR #469, opened 2023-09-07
T343426 Run tests in a headed browser

adeeri:
* To run the tests in a headed browser, set `HEADED_TESTS=true` in `local.env`, similarly to `SELENIUM_LOG_LEVEL`. Then you can watch the test execution with a VNC client on `localhost:7900`.
* Uses https://hub.docker.com/r/selenium/standalone-chrome.
* Running headless is still the default behavior.
* Timings for the whole test run (`make test`) on my machine:
   Headless: 10m50,426s
   Headed: 18m58,860s


lorenjohnson (commented):
Looks great! 

Two things:

1) Definitely currently slows test runs by around ~30% for me from our previous way. Here are the baselines I got just now for `repo` suite:

- `main`:  3:01
- this branch: 4:10
- this branch (`HEADED_TESTS=true`): 04:17

2) I have a preference that we don't break down into a global `test/fixtures` directory for tests data at this point. I would prefer us to map the root of the `test` directory into the browser container as we do with the `test-runner` and use that.

test/suites/wdio.conf.js
```diff
@@ -44,19 +44,19 @@ exports.config = {
...
-		// https://sites.google.com/a/chromium.org/chromedriver/capabilities
 		browserName: 'chrome',
 		maxInstances: 1,
 		'goog:chromeOptions': {
-			// If DISPLAY is set, assume developer asked non-headless or CI with Xvfb.
-			// Otherwise, use --headless (added in Chrome 59)
-			// https://chromium.googlesource.com/chromium/src/+/59.0.3030.0/headless/README.md
 			args: [
-				...( process.env.DISPLAY ? [] : [ '--headless' ] ),
+				...( process.env.HEADED_TESTS ? [ '--disable-gpu' ] : [ '--headless' ] ),
```
RickiJay-WMDE, line 59: Out of curiosity, why is disabling the gpu for headed tests necessary? Is this what's responsible for the slowdown?

lorenjohnson (changes requested):
I'm good with this except three changes:

1) Add `TEST_BROWSER_IMAGE_NAME=selenium/standalone-chrome:116.0.5845.140-chromedriver-116.0.5845.96-grid-4.12.1-20230904` to `test/default.env` and apply that as the image for the `browser` service in  `test/docker-compose.yml` (`image: ${TEST_BROWSER_IMAGE_NAME}). This will allow me to run the native ARM64 community package for the browser and not lose run time :) 

2) Move things that went to `test/fixtures` back to their original location and map the root of `test` (my preference) or `test/suites` into the browser container updating test paths accordingly. 

3) Why the Example test suite failing?

lorenjohnson:
> I'm good with this except three changes:
> 
> 1. Add `TEST_BROWSER_IMAGE_NAME=selenium/standalone-chrome:116.0.5845.140-chromedriver-116.0.5845.96-grid-4.12.1-20230904` to `test/default.env` and apply that as the image for the `browser` service in  `test/docker-compose.yml` (`image: ${TEST_BROWSER_IMAGE_NAME}). This will allow me to run the native ARM64 community package for the browser and not lose run time :)
> 2. Move things that went to `test/fixtures` back to their original location and map the root of `test` (my preference) or `test/suites` into the browser container updating test paths accordingly.
> 3. Why the Example test suite failing?

Actually on #1 we're not currently resolving `test/default.env` into the local shell environment so as proposed will not "just work". We can make a ticket for that change as I think it should be given a moment of thought vs changing to loading those env vars into the local environment. With a ticket created for it I'm fine with that option not being here yet.

lorenjohnson:
So with the ARM build of the browser my `repo` suite test run time was 3:49 (vs ~2:50-3 minutes before) . So this is still at least 25% slower than without the standalone browser, and not a lot better than the emulated browser (which surprises me). Either way though, I'm ok moving forward with this as I do think this setup is tidy and sets us generally the right direction as well as giving us a tidy way to peek in on tests running headed when we need.

lorenjohnson:
Worth noting is you can also watch the headed tests run via Web browser by going to:

http://localhost:7900

...and logging in with the password "secret"

adeeri:
> Worth noting is you can also watch the headed tests run via Web browser by going to:
> 
> http://localhost:7900
> 
> ...and logging in with the password "secret"

Is it not captured in my edit of the README? I didn't mention the password because I guess it depends on the client

lorenjohnson:
> > Worth noting is you can also watch the headed tests run via Web browser by going to:
> > http://localhost:7900
> > ...and logging in with the password "secret"
> 
> Is it not captured in my edit of the README? I didn't mention the password because I guess it depends on the client

Actually I don't think so. This doesn't require a VNC client at all and is a "NoVNC" Webserver instance which runs on Selenium Standalone at http://localhost:7900. So literally you type that address into your Web browser and you can watch live, but you do need to know this "secret" password. That password should be constant as its something configured within the Selenium Standalone container. Does that make sense? 

Give it a shot. I was happy to find this, and it's probably how I'll run headed as its much more convenient that running a VNC client.

test/suites/wdio.conf.js
adeeri: It's actually not necessary -- thanks for catching that :)

rti (changes requested):

rti (approved):
🎈 🎈 🎈 🎈 🎈 🎈

PR #473, opened 2023-09-18
T346023 - Newline Check Not Fix

RickiJay-WMDE:


test/scripts/add_newline.py
```diff
@@ -38,13 +38,23 @@ def add_newline(file: str, root_dir: str) -> bool:
 
         contents = current_file.read(1)
         if not contents == "\n":
-            current_file.write("\n")
+            if should_fix:
+                current_file.write("\n")
+            else:
+                print(f"{file} lacks ending newline")
             return True
```
rti, line 45: It took me quite a while to get that `True` means "there was no newline but it might have been added" and `False` means "there is a newline".

I think with the latest change the logic became quite convoluted, maybe you could make it more obvious by:
- make `add_newline` return `True` if there is a newline, or `should_fix` was given and the newline was successfully added
- return `False` if there is no newline and `should_fix` was not given

This way you can also simplify the check in line 59, because you error exit only if one `add_newline` returned false

test/scripts/add_newline.py
```diff
@@ -38,13 +38,23 @@ def add_newline(file: str, root_dir: str) -> bool:
...
+    should_fix = False
+    if len(sys.argv) > 3:
+        if sys.argv[3] == '--fix':
+            should_fix = True
+    found_errors = False
     for file in sys.argv[2].split("\n"):
         if file_should_be_run(file):
-            add_newline(file, root)
+            found_errors |= add_newline(file, root, should_fix)
+    if found_errors and not should_fix:
```
rti, line 59: see above

test/scripts/lint_fix.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
rti, line 1: Please do not copy code, but make the original `lint.sh` script take a parameter.

rti (changes requested):

test/scripts/lint_fix.sh
RickiJay-WMDE: Implemented, but I haven't worked with parameterized bash scripts before - please double-check me?

rti (approved):

PR #475, opened 2023-09-21
T347002 User executing build does not own the build results (artifacts)

rti:
https://phabricator.wikimedia.org/T347002

This PR makes the builder container run with UID and GID of the system user. So resulting files will be owned by the user calling the builder. Also, the builders' user needs access to the system docker socket, so it gains additional group membership for the system docker group.

This PR draws a harder boundary between host system and builder by not sharing the whole system `/tmp`, but only a custom tmp folder specifically created for individual builder runs (and now also cleaned up afterwards).

During wikibase (bundle) docker builds, a tmp folder for the docker context is created in a tmp dir in order to not change files in the source tree during build. https://phabricator.wikimedia.org/T347053

All the files and folders the builder needs are now mounted, so they are now all listed in `build.sh` instead of some being copied during docker build via the Dockerfile and some being mounted in build.sh. Should help readability.


Dockerfile
```diff
@@ -1,12 +1,13 @@
...
     pip3 install pyyaml
-WORKDIR "/app/"
+
+WORKDIR /app
+
 ENV XDG_CACHE_HOME=/app/cache
-ADD Docker/build/ Docker/build/
-ADD build/ build
+
+RUN mkdir /home/user && chmod 777 /home/user
```
adeeri, line 10: Would it make sense to give the user who executed the build process permissions to the artifacts after they're built (to avoid running as root in the container)?

rti: Thanks for catching the evil 777, applied the change we discussed in our call.

adeeri (approved):

PR #476, opened 2023-09-21
T346210 test_example action is wasteful

rti:
https://phabricator.wikimedia.org/T346210

### DIFF
--- a/.github/workflows/test_example.yml
+++ b/.github/workflows/test_example.yml
@@ -1,13 +1,7 @@
-name: Test Example
+name: Test Example Distribution
 
 on:
-  push:
-    branches:
-      - "main"
-  pull_request:
-    paths-ignore:
-      - 'docs/**'
-      - 'README.md'
+  workflow_dispatch:
 
 jobs:
   test_example:


RickiJay-WMDE (commented):
Under what circumstances does the action now trigger? How do you manually trigger it?

rti:
> Under what circumstances does the action now trigger? How do you manually trigger it?

It can only get triggered manually now. You can do it via the actions UI on the github website.

screenshot-20230925090056

RickiJay-WMDE (approved):

adeeri:
> > Under what circumstances does the action now trigger? How do you manually trigger it?
> 
> It can only get triggered manually now. You can do it via the actions UI on the github website.
> 
> screenshot-20230925090056

If not on every run, when would you want to trigger it?

rti:
> If not on every run, when would you want to trigger it?

This is a very good question. I think we still need a process for that. I'd say directly before we publish a release. I think this will naturally come up again when we discuss our distribution story. I added a note here: https://phabricator.wikimedia.org/T340234

rti:
@adeeri Do you agree merging it?

adeeri:
> @adeeri Do you agree merging it?

Yes, thanks for adding the note so we don't lose it!

adeeri (approved):

PR #478, opened 2023-09-21
T345689 build three versions plain

rti:
In this PR:
- refactor the github action and build script to work similarily for every component (wikibase, wdqs, ...)
- build wikibase (base) and wikibase-bundle with the same make target
- combine github actions so that everything (lint, build, test, publish to GHCR) happens in one workflow

PR #479, opened 2023-09-21
T345689 semver

rti:
# Depends on #478, please review first.

This PR:
- adds tagging built docker images with semver versions configured in the channel they are built for
- refactors some of the build scripts in order to DRY

The images are tagged with the following strategy:

**Prerelease images** get the following tags:
- `service-name-X.Y.Z-pre.I-build-REV-TIMESTAMP-BRANCH` like `wikibase-bundle-2.0.0-pre.1-build-cf84af2-20230926142936-T345689-semver`
- `service-name-X.Y.Z-pre.I` like `wikibase-bundle-2.0.0-pre.1`
- so they do not overwrite any tags that might have been used by previous builds
- so users will never pull those images automatically

**RELEASE images** get the following tags:
- `service-name-X.Y.Z-build-REV-TIMESTAMP-BRANCH` like `wikibase-bundle-2.0.0-build-cf84af2-20230926142936-T345689-semver`
- `service-name-X.Y.Z` like `wikibase-bundle-2.0.0`
- `service-name-X.Y` like `wikibase-bundle-2.0`
- `service-name-X` like `wikibase-bundle-2`
- so they DO overwrite previous builds for the same `MAJOR.MINOR` and `MAJOR` version
- so users might auto update to those builds when they "subscribed" to the respective versions


lorenjohnson:
@rti is this still relevant or will it need to be re-approached now that we've merged #478?

PR #483, opened 2023-09-28
T346740 Avoid reloading images

adeeri:
Checks if an image is already loaded before loading it.


rti (approved):
Cool, that brings a huge speedup! Nice!

PR #480, opened 2023-09-26
T347117 Repo Nuke Test flaky

rti:
https://phabricator.wikimedia.org/T347117

PR #484, opened 2023-09-28
T347117 [BUG] Repo Nuke Test flaky (still)

rti:
#480 did not fix it. We apparently also need to properly wait for the queueing to succeed before we can properly wait for the queue to empty. Makes sense :)


lorenjohnson (approved):
LGTM 👍🏼

PR #485, opened 2023-10-05
T348113 Supress image loading logging

adeeri:
https://phabricator.wikimedia.org/T348113


lorenjohnson (approved):
Yay, I like that better! 👍🏼

PR #488, opened 2023-10-10
Typescript take two

RickiJay-WMDE:


test/helpers/default-functions.ts
```diff
@@ -129,7 +134,7 @@ export function defaultFunctions() {
 		await browser.execute( async () => {
 			const editFormEl = await $( '#editform.mw-editform' );
 			await editFormEl.submit();
-		}, this );
+		} );
```
RickiJay-WMDE, line 137: This was an error typescript alerted - `browser.execute` only expected 1 argument. Please double check.

test/helpers/blazegraph.ts
```diff
@@ -0,0 +1,6 @@
+import Binding from "./types/binding.js";
+
+export function getElementByURI<T>( uri: T, bindings : Binding<T>[] ): Binding<T> | null {
+	const index = bindings.findIndex( ( binding ) => binding.p.value === uri );
+	return index === -1 ? null : bindings[ index ];
```
RickiJay-WMDE, line 5: This previously returned `[]` if `index === -1`. `null` was easier to differentiate and test with in `queryservice-post-upgrade`; some asserts assumed that a binding was returned without checking.

test/helpers/default-functions.ts
```diff
@@ -1,26 +1,31 @@
...
 	 * Returns a Promise
 	 */
-	browser.addCommand( 'makeRequest', ( url, params, postData ) => {
+	browser.addCommand( 'makeRequest', ( url: string, params?: Object, postData?: any ) : Promise<AxiosResponse> => {
 		if ( postData ) {
-			return axios.post( url, postData, params );
+			return axios.default.post( url, postData, params );
 		} else {
-			return axios.get( url, params );
+			return axios.default.get( url, params );
```
RickiJay-WMDE, line 21: Unclear why `axios.default` is needed; will continue looking into that.

test/helpers/types/wdio.d.ts
```diff
@@ -0,0 +1,26 @@
...
+		getDispatchedExternalChange: ( host: string, expectedChange: ExternalChange ) => Promise<ExternalChange | undefined>;
+		getInstalledExtensions: ( server: string ) => Promise<string[] | undefined>;
+		getLuaCpuTime: ( host: string, page: string ) => Promise<LuaCPUValue>;
+		makeRequest: ( url: string, params?: Object, postData?: any ) => Promise<AxiosResponse>;
+		queryBlazeGraphItem: ( itemId: string ) => Promise<Binding[]>;
+		waitForJobs: ( serverURL?: string, timeout?: number, timeoutMsg?: string ) => Promise<boolean>;
+	}
+
+	interface Element {
+	    submit: () => Promise<void>;
```
RickiJay-WMDE, line 24: `Element.submit()` is only used in `editPage` in `default-functions`. Need to check exactly what that thinks it's doing, and perhaps avoid having to define it here.

test/package.json
```diff
@@ -31,8 +33,8 @@
 		"node": ">=10"
 	},
 	"scripts": {
-		"test:run": "wdio ./suites/$SUITE/$SUITE.conf.js",
-		"test:run-filter": "wdio ./suites/$SUITE/$SUITE.conf.js --spec $FILTER",
+		"test:run": "wdio ./suites/$SUITE/$SUITE.conf.ts",
+		"test:run-filter": "wdio ./suites/$SUITE/$SUITE.conf.ts --spec $FILTER",
```
RickiJay-WMDE, line 37: Turns out you can fully just run these with Typescript without explicitly compiling them to Javascript first

test/wdio.conf.ts
```diff
@@ -8,12 +8,13 @@ import JsonReporter from './helpers/json-reporter.js';
 import { defaultFunctions as defaultFunctionsInit } from './helpers/default-functions.js';
 import saveScreenshot from './helpers/WDIOMediawikiScreenshotPatch.js';
 import WikibaseApi from './helpers/WDIOWikibaseApiPatch.js';
+import {Options, Frameworks} from "@wdio/types"
```
lorenjohnson, line 11: I would have expected to see spaces surrounding this object expansion. Is it your expectation to not have them, and is our linting specific to not having spacing wrap object expansions or agnostic?

test/helpers/json-reporter.ts
```diff
@@ -0,0 +1,68 @@
+import { existsSync, readFileSync, writeFileSync } from 'fs';
+import reporter, { SuiteStats, TestStats } from '@wdio/reporter';
```
lorenjohnson, line 2: I think `reporter` should be capitalised

lorenjohnson (changes requested):
Looks great! Good work! Would like to see the WDIO `reporter` class capitalised on import and use (see comment), otherwise good to go in my book 👍🏼

test/helpers/json-reporter.ts
RickiJay-WMDE: Updated!

test/wdio.conf.ts
RickiJay-WMDE: `eslint` as currently configured doesn't format Typescript files. I had intended to ask if a fix for that should be part of this ticket or a folo?

lorenjohnson: Yes, I do think that would be best as part of this. Are you ok adding that, or does it come with new challenges? Also, I think are we using some sort of WikiMedia standard linting config and if so, we need to see if there is a Typescript specific config we should inherit?

RickiJay-WMDE: So... this has quite a few challenges. There's a project linked below for `eslint-config-wikimedia-typescript`; it hasn't been officially adopted yet, and doesn't seem to work with `wikimedia/selenium`. I copied the `selenium.json` and `server.json` files from eslint-config-wikimedia (linked below), edited them to remove `wikimedia/language/es2018` from the latter, and disabled these rules:

```
       "jsdoc/check-param-names": "off",
        "jsdoc/require-param": "off",
        "no-undef": "off",
        "node/no-missing-import": "off"
```

I'm meeting tomorrow with Michael Große, who worked on the typescript config project, to see if there's a simpler implementation.

https://github.com/wmde/eslint-config-wikimedia-typescript

https://github.com/wikimedia/eslint-config-wikimedia/tree/main

RickiJay-WMDE: Updates: Found an easier way!

lorenjohnson: Nice. Was hoping this would be what came next 🙃

lorenjohnson (approved):
Looks good and tested locally. Everything is happy 🎉

PR #490, opened 2023-10-16
T346214 await displayed

RickiJay-WMDE:
### `awaitDisplayed`

If a string is passed, await finding that element. If a promise of an element is passed, await the element. In any case, wait until the element is displayed. Return the element.

### `SubmittablePage`

A page with a `submit()` method and a `submitBtn`, inheritable. We had the exact same `submit()` method implemented thrice.


rti (approved):
💜💜💜 nice work. Thanks a lot!

PR #494, opened 2023-10-26
`wdio-mediawiki 2.4.0` - Remove Local Patches

RickiJay-WMDE:
The code changes we've made to `wdio-mediawiki` will eliminate the need for our patches; this updates to `wdio-mediawiki` `2.4` and removes our local patch files.

The foundation is planning on reverting our changes as, I believe, `2.5.0` as soon as I indicate that `2.4` works for us, and they're taking a different approach when they go to `wdio v8` - they don't want to be backwards compatible with `wdio v7`, so that'll be `wdio-mediawiki` `3.0.0`. We'll need to NOT update to `2.5`, and update our code to be compatible with theirs when they publish `3.0`.


rti (approved):
What a journey! Congratz!! Approved!

PR #506, opened 2023-11-03
T350363 - Remove unnecessary `waitForDisplayed()`

RickiJay-WMDE:
https://phabricator.wikimedia.org/T350363


lorenjohnson:
@RickiJay-WMDE I like this approach. I started by just keeping `awaitDisplayed` in place and commenting out the `waitForDisplayed` calls within it. In local runs I was finding some tests right away fail. Surprisingly, it looks like all the tests passed here on CI without any `waitForDisplayed` call.

Either way, I much prefer just removing the helper as you've done here, and I'll continue off of this branch and do test runs, adding back `waitForDisplayed` in each case where the tests fail locally.

RickiJay-WMDE:
Hmmm. special-version and special-property are failing for me now, are they working for you?

lorenjohnson:
> Hmmm. special-version and special-property are failing for me now, are they working for you? @lorenjohnson

Yes, everything was passing on multiple runs but with the merge up of `main` need to re-run. Will report back.

lorenjohnson (approved):
LGTM

PR #507, opened 2023-11-08
chore(release): wmde.14

rti:
https://phabricator.wikimedia.org/T340939

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,18 @@ This file is only intended to serve the developers of this repository and of the
 
 This provides an overview of the releases that have been made using this release pipeline.
 
+
+## November 2023: First 1.40 release
+
+The [standard MediaWiki upgrade process](https://www.mediawiki.org/wiki/Manual:Upgrading) can be followed for this release.
+
+- [MediaWiki release notes](https://www.mediawiki.org/wiki/Release_notes/1.40)
+- [Wikibase release notes](https://github.com/wikimedia/Wikibase/blob/REL1_40/RELEASE-NOTES-1.40)
+
+| Suite Version | Version Info | Date available | Run number | Release task  |
+|---------------|----------------------|----------------|------------| --------------|
+| wmde.14 ([spec](https://github.com/wmde/wikibase-release-pipeline/blob/wmde.14/variables.env)) | MediaWiki (1.40.1, PHP 8.1.24), WDQS (0.3.135)    | 10 November 2023 | [TODO](https://github.com/wmde/wikibase-release-pipeline/actions/runs/TODO) | [T340939](https://phabricator.wikimedia.org/T340939) |
+
 ## October 2023: Security releases for 1.39.5
 
 - [Maintenance release: MediaWiki 1.39.2](https://lists.wikimedia.org/hyperkitty/list/wikitech-l@lists.wikimedia.org/message/UJBH5OGW5M5Q22KETIQY4MV2Q6XJPQWG/)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,9 @@ This provides an overview of the releases that have been made using this release
 
 ## November 2023: First 1.40 release
 
+- [MediaWiki release announcement 1.40](https://lists.wikimedia.org/hyperkitty/list/wikitech-l@lists.wikimedia.org/message/H46H5ZYZG2PYUQ5STK7NWKF7GXYW7H6B/)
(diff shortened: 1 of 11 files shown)


CHANGES.md
```diff
@@ -4,6 +4,18 @@ This file is only intended to serve the developers of this repository and of the
...
+## November 2023: First 1.40 release
+
+The [standard MediaWiki upgrade process](https://www.mediawiki.org/wiki/Manual:Upgrading) can be followed for this release.
+
+- [MediaWiki release notes](https://www.mediawiki.org/wiki/Release_notes/1.40)
+- [Wikibase release notes](https://github.com/wikimedia/Wikibase/blob/REL1_40/RELEASE-NOTES-1.40)
+
+| Suite Version | Version Info | Date available | Run number | Release task  |
+|---------------|----------------------|----------------|------------| --------------|
+| wmde.14 ([spec](https://github.com/wmde/wikibase-release-pipeline/blob/wmde.14/variables.env)) | MediaWiki (1.40.1, PHP 8.1.24), WDQS (0.3.135)    | 10 November 2023 | [TODO](https://github.com/wmde/wikibase-release-pipeline/actions/runs/TODO) | [T340939](https://phabricator.wikimedia.org/T340939) |
```
darthmon: In the last release we added the apache version.  
I just checked and it's again: debian/apache2 2.4.57-2

(see https://hub.docker.com/layers/library/mediawiki/latest/images/sha256-8461b941852c8a22475753f49971f8d4b0b452f4ef3bf979912e291f9b9767fb?context=explore for more info)

darthmon: is there a reason why you left it out?

darthmon: This may have changed through the new branches system but I will ask because I cannot find it:
https://github.com/wmde/wikibase-release-pipeline/blob/wmde.14/variables.env

this is the old wmde.14/versions/wmde14.env?

variables.env
```diff
@@ -10,28 +10,28 @@ MEDIAWIKI_VERSION=1.40.1
...
-# Head of master on 2023-10-11
+# Head of master on 2023-11-09
 # as per https://bitbucket.org/magnusmanske/magnustools/commits/branch/master
-MAGNUSTOOLS_COMMIT_HASH=11e19400271a47c4d7e927bcb7aa95f1bfaf38a9
+MAGNUSTOOLS_COMMIT_HASH=e4abdb6b3bedfdde0325df513b272e1972b1af36
 
-# Head of master on 2023-10-11
+# Head of master on 2023-11-09
 # as per https://github.com/ProfessionalWiki/WikibaseLocalMedia/tree/master
 WIKIBASELOCALMEDIA_COMMIT_HASH=6dcc6e82526896352b6bae1df11750f378201fcb
```
darthmon: this is surely right since it is the parent commit (https://github.com/ProfessionalWiki/WikibaseLocalMedia/commit/6dcc6e82526896352b6bae1df11750f378201fcb) to yours.

I would like to know though why we are not using the commit from the patch that you had to add on that code base, which is 74c3f70893ffd76feefa9f1e3d77052db20ec3ac?

variables.env
```diff
@@ -10,28 +10,28 @@ MEDIAWIKI_VERSION=1.40.1
 RELEASE_MAJOR_VERSION=1.40
 RELEASE_VERSION=1.40.1
 
-# WDQS_VERSION 0.3.135 is latest release on 2023-10-11
+# WDQS_VERSION 0.3.136 is latest release on 2023-11-09
 # as per https://github.com/wikimedia/wikidata-query-rdf/tags
 # and https://archiva.wikimedia.org/repository/releases/org/wikidata/query/rdf/service/
 WDQS_VERSION=0.3.135
```
darthmon, line 16: just checking: we are staying on the 0.3.135 even if there is a newer one, right? why?

CHANGES.md
rti: No, thanks for checking.
Added.

rti: Good catch. This indeed changed with the branch system. This is the path where the correct file (the one replacing `versions/wmde14.env`) should end up after the merge of this PR. So it should work 🤞🏾

variables.env
rti: This has been changed, for some reason you did not see the change in your review. Did I forgot to push something? Not it is there https://github.com/wmde/wikibase-release-pipeline/pull/507/files#diff-3aa2b6a722522c6a53ed1ab4f2d690408c5fbf68597868a4e331fd13c4bc0933R30

variables.env
rti: This was the version I did the product acceptance tests with. If we bumped the WDQS version now, I think we would need to do the product acceptance tests again. But you are right, I should have bumped the version before the product acceptance tests. This was an oversight on my side.

PR #508, opened 2023-11-08
fix: change upgrade test 1.39 base version to latest

rti:
Now, as wmde.13 is released, we should test updates on main from there.


lorenjohnson (approved):
🌱

PR #513, opened 2023-11-14
T350879: Special Property: Item

RickiJay-WMDE:
Existing test creates a property of type "String". [The testing spreadsheet](https://docs.google.com/spreadsheets/d/1VNT1DZcH3LREhduS5Zsa9jy-esffwNv9potBylzkZC8/edit#gid=0) indicates a property of type "Item".

Interestingly, while you could pre-supply "string" as a type, you cannot do the same with "Item", so I had to make the test click the type dropdown and select "Item". Which is what the user would do, so.


test/specs/repo/special-property.ts
```diff
@@ -37,4 +37,25 @@ describe( 'Special:NewProperty', function () {
...
+
+		await SpecialNewProperty.labelInput.setValue( 'Cool Item label' );
+		await SpecialNewProperty.descriptionInput.setValue( 'Cool Item description' );
+		await SpecialNewProperty.aliasesInput.setValue(
+			'Great Item!|Greatest Item!'
+		);
+
+		await SpecialNewProperty.datatypeInput.click();
+		await browser.pause( 1 * 1000 );
+		await $( '.oo-ui-labelElement-label=Item' ).click();
```
rti, line 52: You can select the *Item* datatype with `wiki/Special:NewProperty?datatype=wikibase-item`.

I like the idea to actually use the UI to select the datatype, I do not like, that we add another wait for that. I'd suggest:
 - either just use the url again
 - or find a way to properly wait for the element you want to click on

rti (changes requested):

test/specs/repo/special-property.ts
lorenjohnson: Yes, unless unless there is no other way, I think we shouldn't be using `browser.pause`. If we can do that through UI selection and await display of a UI element I also agree that would be best. 

In this case I would DRY this one up and use the existing "Should be able to add statement and reference to property" test iterate over `[ 'string', 'wikibase-item' ]`. Bonus: add another type or two to the array.

RickiJay-WMDE: You could use the url, yes, but I'm trying to match the test instructions, which do include clicking on the relevant type
I'll see about removing the pause

RickiJay-WMDE: Didn't update here, apparently, but I removed the pause

rti (approved):

PR #528, opened 2023-11-21
T350879 extend special item test

adeeri:
Verify that the created page's contents are as expected, and not only that the page exists.


rti (approved):
LGTM 👍🏾

PR #529, opened 2023-11-21
chore: bump version for next release

rti:
As `wmde.14` branched of, we should change the version in main. Actually, I think we should have done this right after branching off `wmde.14`.


lorenjohnson (approved):
Agreed...

PR #533, opened 2023-12-04
T350879: Item Properties

RickiJay-WMDE:
"Add a statement to a property (data type: item)" from the [testing spreadsheet](https://docs.google.com/spreadsheets/d/1VNT1DZcH3LREhduS5Zsa9jy-esffwNv9potBylzkZC8/edit#gid=0)

Run the `property` tests on wikibase-items as well as strings. Clarify creation, check, creation, check, and split into specific tests.


lorenjohnson (commented):
This looks good but would like to have a session to talk about: 1) the new TS types and how we're organising those, and, 2) the use of `browser.pause()` and if you have explored how it might be avoided here (for instance through `browser.waitForJobs()`). I would prefer to do that session on video but we could chat here too if it takes too long to get to that :)

lorenjohnson (approved):

PR #525, opened 2023-11-21
OLD - T350782 Test setup done in JS / WDIO vs Shell scripts

lorenjohnson:
Makes possible running tests more atomically, and more. https://phabricator.wikimedia.org/T350782

* Replaces all test runner shell scripts and with Typescript making more clear and flexible control of the our test environment
* Update test suite setup to allow for a full set of overrides to the test setup, and running process. See `*.conf.js` files and review the related Typescript types which self-document much of what is happening. 
* Provides detailed WDIO test run logs in `suites/SUITE_NAME/results/wdio*.log`. Here we can see all browser console output, as well as detail results of each selector happening in each spec file. Setup log from steps before test runners start is `wdio.log`, subsequent log files are created for each spec file ran according to the correlated WDIO job runner number, e.g. `wdio-0-1.log`, etc. 
* Changes local dev test run CLI from `make test SUITE=<suite-name>` to `./test.sh <suite-name>`
* Adds `--setup` test runner option to bring-up a test suites services without running tests, and leave them up along with the Selenium instance for manual testing, e.g `./test.sh quickstatements --setup`
* Allows all WDIO Runner options to be added to test runs :`./test.sh <suite-name> <WDIO-Runner-CLI-options>`. E.g. `./test.sh repo --spec=specs/repo/property.ts --watch`, etc
* Rewrites Example test setup to be in line with other tests. The Example test suite is now ran using simply: `./test.sh example`
* Rewrites Upgrade test setup to be in line with other tests. Run the Upgrade test suite with: `./test.sh upgrade WMDE9 <optional-version-tag-of-upgrade-to-version>`
* Renames all `*_SERVER` env variables to `*_URL` in the test environment (the only place we were using this variables) and adds the `http://` protocol prefix where it wasn't consistently applied before.
* Removes linting run before test runs
* Fix: Removes password requirement on Selenium Grid / Test Browser instance (docs updated)
* Fix: Linting no longer tries to lint Dockerfiles within `node_modules` directories
* Fix: In Github Actions removes `lint` as a "need" of the `build` and `test` jobs . This will have the effect of the checks still failing if linting fails, but `build` and `test` continue running. 
... etc

---

Please try out the new Test CLI, here is the help for that which can be retrieved by simply typing `./test.sh` or `./test.sh --help`:

```
> ./test.sh --help

./test.sh <suiteName>

Run a test suite

Commands:
  ./test.sh <suiteName>                        Run a test suite                                                [default]
[...]

PR #542, opened 2023-12-16
T350782 Test System Rewrite

lorenjohnson:
Replaces PR #525 with a much more clear commit history. This branch also no longer includes linting of `docs/diagrams` or the property test updates.

---

Makes possible running tests more atomically, and more. https://phabricator.wikimedia.org/T350782

* Replaces all test runner shell scripts and with Typescript making more clear and flexible control of the our test environment
* Update test suite setup to allow for a full set of overrides to the test setup, and running process. See `*.conf.js` files and review the related Typescript types which self-document much of what is happening. 
* Provides detailed WDIO test run logs in `suites/SUITE_NAME/results/wdio*.log`. Here we can see all browser console output, as well as detail results of each selector happening in each spec file. Setup log from steps before test runners start is `wdio.log`, subsequent log files are created for each spec file ran according to the correlated WDIO job runner number, e.g. `wdio-0-1.log`, etc. 
* Changes local dev test run CLI from `make test SUITE=<suite-name>` to `./test.sh <suite-name>`
* Adds `--setup` test runner option to bring-up a test suites services without running tests, and leave them up along with the Selenium instance for manual testing, e.g `./test.sh quickstatements --setup`
* Allows all WDIO Runner options to be added to test runs :`./test.sh <suite-name> <WDIO-Runner-CLI-options>`. E.g. `./test.sh repo --spec=specs/repo/property.ts --watch`, etc
* Rewrites Example test setup to be in line with other tests. The Example test suite is now ran using simply: `./test.sh example`
* Rewrites Upgrade test setup to be in line with other tests. Run the Upgrade test suite with: `./test.sh upgrade WMDE9 <optional-version-tag-of-upgrade-to-version>`
* Renames all `*_SERVER` env variables to `*_URL` in the test environment (the only place we were using these variables) and adds the `http://` protocol prefix where it wasn't consistently applied before.
* Removes linting run before test runs
* Fix: Removes password requirement on Selenium Grid / Test Browser instance (docs updated)
* Fix: Linting no longer tries to lint Dockerfiles within `node_modules` directories
* Fix: In Github Actions removes `lint` as a "need" of the `build` and `test` jobs . This will have the effect of the checks still failing if linting fails, but `build` and `test` continue running and building runs in parallel with linting decreasing test run times by a couple minutes.
... etc

---

Please try out the new Test CLI, here is the help for that which can be retrieved by simply typing `./test.sh` or `./test.sh --help`:

```
> ./test.sh --help

./test.sh <suiteName>

Run a test suite

Commands:
  ./test.sh <suiteName>                        Run a test suite                                                [default]
  ./test.sh upgrade <fromVersion> [toVersion]  run upgrade test
  ./test.sh all                                Run all test suites

Positionals:
  suiteName  Name of the test suite to run or setup
   [string] [choices: "repo", "fedprops", "repo_client", "quickstatements", "pingback", "confirm_edit", "elasticsearch",
                                       "base__repo", "base__repo_client", "base__pingback", "base__fedprops", "example"]

Options:
  --setup                  Start and leave-up the test environment without running tests                       [boolean]
  --headedTests, --headed  Run tests in a headed browser                                                       [boolean]
  --spec                   Run only this spec file(s) using the selected test suite setup (WDIO)                 [array]
  --watch                  Automatically re-run tests when spec files in the suite change (WDIO)               [boolean]
  --version                Show version number                                                                 [boolean]
  --help                   Show help                                                                           [boolean]
```

Also help is available specific to the Upgrade test suite:

```
> ./test.sh upgrade --help

./test.sh upgrade <fromVersion> [toVersion]

run upgrade test

Positionals:
  fromVersion  Version to upgrade FROM in the upgrade test
  [string] [required] [choices: "WMDE9", "WMDE9_BUNDLE", "WMDE10", "WMDE10_BUNDLE", "WMDE11", "WMDE11_BUNDLE", "WMDE12",
                                        "WMDE12_BUNDLE", "WMDE13", "WMDE13_BUNDLE", "LOCAL_BUILD", "LOCAL_BUILD_BUNDLE"]
  toVersion    Version to upgrade TO in the upgrade test
  [string] [choices: "WMDE9", "WMDE9_BUNDLE", "WMDE10", "WMDE10_BUNDLE", "WMDE11", "WMDE11_BUNDLE", "WMDE12", "WMDE12_BU
                                                  NDLE", "WMDE13", "WMDE13_BUNDLE", "LOCAL_BUILD", "LOCAL_BUILD_BUNDLE"]

Options:
  --setup                  Start and leave-up the test environment without running tests                       [boolean]
  --headedTests, --headed  Run tests in a headed browser                                                       [boolean]
  --spec                   Run only this spec file(s) using the selected test suite setup (WDIO)                 [array]
  --watch                  Automatically re-run tests when spec files in the suite change (WDIO)               [boolean]
  --version                Show version number                                                                 [boolean]
  --help                   Show help                                                                           [boolean]
```


test/test-runner.env
```diff
@@ -0,0 +1,8 @@
+# You can override any of these in ../../local.env
+HEADED_TESTS=false
+MOCHA_OPTS_TIMEOUT=120000
```
lorenjohnson, line 3: This is an increase from the previous default of 90 seconds which was often not enough in CI to get through the Example test suite. WIth some additional elaboration we could set a different timeout for CI vs local, but this extra 30 seconds of timeout hasn't hurt me locally.

test/test-services.env
```diff
@@ -0,0 +1,60 @@
+# Any of these can be overridden locally in ../local.env
+
+# == TEST SERVICES
+
+# wikibase / wikibase-client / wikibase-jobrunner
+WIKIBASE_URL=http://wikibase.svc
```
lorenjohnson, line 6: `*_SERVER` URLs all are now named `*_URL` and always have a protocol in front of them. These env vars are only used in the test environment. * In this case `MW_` is also renamed to `WIKIBASE_` which I find a bit more explicit in our context.

lint.sh
```diff
@@ -1,6 +1,14 @@
 #!/usr/bin/env bash
 set -e
 
+# ℹ️ Linting Dockerfiles (**/Dockerfile)
+# https://github.com/hadolint/hadolint
+docker run --rm -v "$(pwd)":/code -v "$(pwd)/.hadolint.yml":/.hadolint.yml hadolint/hadolint:latest-alpine sh -c "find . -name Dockerfile -print -o -type d -name node_modules -prune | xargs hadolint"
```
lorenjohnson, line 6: No longer tries to lint `Dockerfile`s installed in node dependencies (i.e. ignores `node_modules`)

lint.sh
```diff
@@ -29,8 +36,14 @@ docker run --rm -v "$(pwd)":/code -v "$(pwd)/.hadolint.yml":/.hadolint.yml hadol
 # https://github.com/koalaman/shellcheck#from-your-terminal
 find . -type d -name node_modules -prune -false -o -name "*.sh" -print0 | xargs -0 docker run --rm -v "$(pwd)":/code dcycle/shell-lint:2
 
+if ! [[ -f "local.env" ]]; then
+	touch local.env
```
lorenjohnson, line 40: If `local.env` doesn't exist Docker will complain when the test-runner is ran (which is used for linting as well as test runs).

test/Dockerfile
```diff
@@ -1,18 +1,26 @@
 FROM node:18.17.1-bullseye
 
+ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true
```
lorenjohnson, line 3: The default installation of Chromium in the test-runner image is no longer needed, and causes issues on MacOS.

test/cli.js
```diff
@@ -0,0 +1,165 @@
+import { Launcher } from '@wdio/cli';
```
lorenjohnson, line 1: This one can be a bit hard to understand without digging into https://www.npmjs.com/package/yargs. Type `./test.sh` when on this branch to see help for the CLI options this provides.

test/helpers/default-functions.ts
```diff
@@ -70,52 +84,28 @@ export function defaultFunctions(): void {
...
 			);
-			return lodash.map( result.data.query.extensions, 'name' );
-		}
-	);
-
-	/**
-	 * Execute docker command on container and get output
-	 */
-	browser.addCommand(
-		'dockerExecute',
```
lorenjohnson, line 89: This command is now handled by `TestEnv#runDockerComposeCmd` as you will see applied in the `dbQuery` command above (the only place it was used). `TestEnv` is now where all Docker test service things are handled.

test/helpers/default-functions.ts
```diff
@@ -31,12 +46,12 @@ export function defaultFunctions(): void {
...
 	browser.addCommand(
 		'dbQuery',
-		async ( query: string, config?: DatabaseConfig ) => {
+		( query: string, config?: DatabaseConfig ): string => {
 			if ( !config ) {
 				config = {
-					user: process.env.DB_USER,
-					pass: process.env.DB_PASS,
-					database: process.env.DB_NAME
+					user: testEnv.vars.DB_USER,
```
lorenjohnson, line 52: Note that in most all cases where `process.env.*` was used in the past you should now use `testEnv.vars` where the env vars related to the services being tested are now loaded (by default from `variables.env`, `test/test-services.env`, and `local.env` -- in that order). Things related to the WDIO test runner, such as timeouts and log levels, are set in `test/test-runner.env` and referenced directly from `process.env`.

test/helpers/default-functions.ts
```diff
@@ -11,17 +11,32 @@ import ExternalChange from '../types/external-change.js';
...
-		( url: string, params?: any, postData?: any ): Promise<AxiosResponse> => {
+		async (
+			url: string,
+			params?: Record<string, unknown>,
+			postData?: Record<string, unknown>
+		): Promise<Partial<AxiosResponse>> => {
 			if ( postData ) {
-				return axios.post( url, postData, params );
+				const { data, status } = await axios.post( url, postData, params );
+				return { data, status };
```
lorenjohnson, line 36: This change eliminates the entire Axios Response object from being logged.

test/setup/loadLocalDockerImage.ts
```diff
@@ -0,0 +1,36 @@
+import { SpawnSyncReturns, spawnSync } from 'child_process';
+
+// https://regex101.com/r/eK9lPd/3
+// eslint-disable-next-line security/detect-unsafe-regex
+export const dockerImageUrlRegExp = /^(?<Name>(?<=^)(?:(?<Domain>(?:(?:localhost|[\w-]+(?:\.[\w-]+)+)(?::\d+)?)|[\w]+:\d+)\/)?\/?(?<Namespace>(?:(?:[a-z0-9]+(?:(?:[._]|__|[-]*)[a-z0-9]+)*)\/)*)(?<Repo>[a-z0-9-]+))[:@]?(?<Reference>(?<=:)(?<Tag>[\w][\w.-]{0,127})|(?<=@)(?<Digest>[A-Za-z][A-Za-z0-9]*(?:[-_+.][A-Za-z][A-Za-z0-9]*)*[:][0-9A-Fa-f]{32,}))?/;
```
lorenjohnson, line 5: This is a transcription of Docker's own URL parsing RegEx. For our purposes this could be done in a more naive way, but I think this is going to serve us better. The regex101.com link in the comment tracks changes and any issues, as well as it gives a sandbox for testing the RegEx.

test/setup/wdio.conf.ts
```diff
@@ -0,0 +1,141 @@
+/* eslint camelcase: ["error", {properties: "never"}] */
```
lorenjohnson, line 1: Most values here should be derived from the currently globally set `TestEnv` and not set directly in this file. The one exception is `capabilities` (i.e. our Selenium Grid Standalone / Test Browser setup) which largely is configured only here.

test/specs/repo_client/extensions/scribunto-item.ts
```diff
@@ -3,7 +3,6 @@ import assert from 'assert';
 import LoginPage from 'wdio-mediawiki/LoginPage.js';
 import { stringify } from 'querystring';
 import { readFile } from 'fs/promises';
-import { skipIfExtensionNotPresent } from '../../../helpers/default-functions.js';
```
lorenjohnson, line 6: This command has now been added to `browser`

test/specs/upgrade/upgrade.ts
```diff
@@ -0,0 +1,89 @@
+import { spawnSync } from 'child_process';
+import WikibaseApi from 'wdio-wikibase/wikibase.api.js';
+import { versions } from '../../suites/upgrade/versions.js';
+import { getTestString } from 'wdio-mediawiki/Util.js';
+import assert from 'assert';
+
+describe( 'Wikibase upgrade', function () {
+	let oldItemID: string;
+
+	before( async () => {
```
lorenjohnson, line 10: This before hook works in concert with what happens in `test/suites/upgrade/upgrade.conf.ts`. See comments.

test/suites/base__repo/base__repo.conf.ts
```diff
@@ -1,12 +1,21 @@
-import { config as wdioConf } from '../../wdio.conf.js';
+import TestEnv from '../../setup/TestEnv.js';
+import { defaultTestSettings } from '../../setup/makeTestSettings.js';
+import wdioConfig from '../../setup/wdio.conf.js';
 
-export const config: WebdriverIO.Config = {
-	...wdioConf,
+export const testEnv = TestEnv.createWithDefaults( {
+	name: 'base__repo',
+	isBaseSuite: true,
```
lorenjohnson, line 7: This can be inferred from the name of the test suite, but I chose to make it more explicit for clarity and flexibility. This flag is only used in the `beforeServices` hook in choosing which Docker images to load from the local build.

test/docker-compose.yml
```diff
@@ -1,59 +1,42 @@
...
       dockerfile: Dockerfile
+    stop_grace_period: 1s
     entrypoint: /bin/bash
     volumes:
-      - .:/usr/src/app
-      - node_modules:/usr/src/app/node_modules
+      - ..:/usr/src
+      - node_modules:/usr/src/test/node_modules
       - /var/run/docker.sock:/var/run/docker.sock
+    env_file: ./test-runner.env
```
lorenjohnson, line 22: This makes all env vars in this file directly available in the test-runner container.

README.md
```diff
@@ -34,10 +34,10 @@ $ ./build.sh --no-cache wdqs
 ### Test Commands
 
 ```
-$ make test
-$ make test SUITE=repo
-$ make test SUITE=repo FILTER=special-item
-$ make test-upgrade VERSION=wmde.9
+$ ./test.sh
+$ ./test.sh repo
+$ ./test.sh repo --spec special-item
```
RickiJay-WMDE, line 39: Please update to note the need for path, extension, and the root of the path

RickiJay-WMDE (changes requested):

test/test-services.env
RickiJay-WMDE: Is this then a breaking change?

lorenjohnson: Nope, these variables are only used within the test environment and have no impacted outside of it (e.g. in our Build setup, Docker images, or Example setup)

README.md
RickiJay-WMDE: As a follow-up, we might want to see if there's a way to default that to specs/SUITE/ as the root?

test/setup/loadEnvFiles.ts
```diff
@@ -0,0 +1,36 @@
...
+		processEnv: providedEnvvars
+	} );
+
+	if ( result.error ) {
+		throw result.error;
+	}
+
+	const envVarsResult = dotenvExpand.expand( {
+		parsed: providedEnvvars, ignoreProcessEnv: true
+	} );
```
RickiJay-WMDE, line 20: Framing is confusing - it looks like you're returning the variables from the config file into `results`, and then failing to include that in `dotenvExpand.expand`. Would just fetching `{ error }` instead of `result` work?

RickiJay-WMDE (changes requested):

test/setup/loadEnvFiles.ts
RickiJay-WMDE: Naming the variable `providedEnvvars` is part of the issue, I think - the variable name does not suggest it to be mutable.

test/setup/loadLocalDockerImage.ts
```diff
@@ -0,0 +1,36 @@
...
+export const loadLocalDockerImage = (
+	dockerImageURL: string,
+	reload: boolean = false
+): SpawnSyncReturns<string> => {
+	try {
+		const dockerImageUrlMatch = dockerImageUrlRegExp.exec( dockerImageURL );
+
+		if ( dockerImageUrlMatch.groups.Repo ) {
+			const imageName = dockerImageUrlMatch.groups.Repo;
+			const { stdout: alreadyLoaded } = spawnSync( 'docker',
```
RickiJay-WMDE, line 16: This is called within the `async` `beforeServices`, should it be `sync`?

test/setup/makeTestSettings.ts
```diff
@@ -0,0 +1,124 @@
...
+		'../local.env'
+	],
+	composeFiles: [
+		'suites/docker-compose.yml'
+	],
+	waitForUrls: (): string[] => {
+		return ( [
+			`${testEnv.vars.WIKIBASE_URL}/wiki/Main_Page`,
+			`${testEnv.vars.WDQS_URL}/bigdata/namespace/wdq/sparql`,
+			testEnv.vars.WDQS_FRONTEND_URL
```
RickiJay-WMDE, line 28: Isn't using `testEnv` in `makeTestSettings` circular?

test/setup/wdio.conf.ts
```diff
@@ -0,0 +1,141 @@
...
+			{
+				browserName: 'chrome',
+				maxInstances: settings.maxInstances,
+				'goog:chromeOptions': {
+					args: [
+						// The window size is relevant for responsive pages rendering differently on
+						// different screen sizes. Bootstrap considers widths between 1200 and 1400
+						// as XL, let's use that.
+						// https://getbootstrap.com/docs/5.0/layout/breakpoints/#available-breakpoints
+						...[ '--window-size=1280,800' ],
```
RickiJay-WMDE, line 41: Please just use `'--window-size=1280,800'` instead of `...[ '--window-size=1280,800' ]`

RickiJay-WMDE (changes requested):

README.md
lorenjohnson: Why using a search string (i.e. "property") vs the path to the spec didn't work is actually a known issue on WDIO 8 which was resolved in August, but we're not currently up enough minor/patch version to get it. Here is the related Github issue: https://github.com/webdriverio/webdriverio/issues/9495

It's equally of issue in our current test system using the `FILTER=` env var. Does `make test SUITE=repo FILTER=property` work for you on main either? It doesn't currently for me. 

Otherwise, I actually want the file path to be explicit so that everything in test always has the same file root for simplicity. These are the same paths we us in the `conf.ts` files for pointing to spec files. Both settings are direct passthroughs to WDIO so they work the same and the WDIO docs are valid. That all makes for me less cognitive dissonance. Also this is modeled in such a way thar any spec/s can be ran against any test env setup and the two aren't unnecessarily coupled.

test/setup/makeTestSettings.ts
lorenjohnson: They are within a function definition and not evaluated until the 'testEnv` is created and running and we check if the services are up in `TestEnv#up` (which is ran by WDIO `onPrepare` hook). The functions are just being defined, not evaluated. It's the same as using `browser.addCommand` in `default-functions.ts`, as `browser` doesn't exist either until the `before` hook is ran.

test/setup/wdio.conf.ts
lorenjohnson: For sure, I can and agree, but this is current code, nothing I changed.

test/setup/loadEnvFiles.ts
lorenjohnson: Certainly you have it right that `provided*` var naming for me is always chosen to indicate immutability. So will change that and see what can be done to clarify things here more.

The two libraries I'm using here work fairly implicitly and their APIs kinda suck, so I blame that a bit.

flake.nix
```diff
@@ -34,10 +34,10 @@
             echo "$ ./build.sh --no-cache wdqs"
             echo
             echo "Test Commands:"
-            echo "$ make test"
-            echo "$ make test SUITE=repo"
-            echo "$ make test SUITE=repo FILTER=special-item"
-            echo "$ make test-upgrade VERSION=wmde.9"
+            echo "$ ./test.sh"
+            echo "$ ./test.sh repo"
+            echo "$ ./test.sh repo --spec special-item"
```
RickiJay-WMDE, line 39: Same as other spec documentation

test/specs/repo/extensions/babel.ts
```diff
@@ -1,18 +1,16 @@
 import assert from 'assert';
 import LoginPage from 'wdio-mediawiki/LoginPage.js';
-import { skipIfExtensionNotPresent } from '../../../helpers/default-functions.js';
-
```
RickiJay-WMDE, line 4: Spacing

test/specs/pingback/pingback.ts
```diff
@@ -14,7 +14,7 @@ describe( 'Pingback', function () {
 		assert.strictEqual( sqlResult.includes( 'WikibasePingback-1.' ), true );
 
 		const result = await browser.makeRequest(
-			process.env.PINGBACK_BEACON_SERVER
+			'http://mediawiki.svc'
```
RickiJay-WMDE, line 17: Any particular reason not to keep this in a variable?

test/helpers/default-functions.ts
```diff
@@ -70,52 +84,28 @@ export function defaultFunctions(): void {
...
-		async ( server: string ): Promise<string[] | undefined> => {
-			const result = await browser.makeRequest(
-				`${server}/w/api.php?action=query&meta=siteinfo&siprop=extensions&format=json`
+		'skipIfExtensionNotPresent',
+		async (
+			test: Context,
+			extension: string
+		): Promise<void> => {
+			const installedExtensions = await getInstalledExtensions(
+				testEnv.vars.WIKIBASE_URL
```
RickiJay-WMDE, line 96: At the risk of being annoying
The switch from `getInstalledExtensions` to `skipIfExtensionNotPresent` does seem like a discrete, separable change
Thoughts on splitting that out?

lorenjohnson: This is actually required by this changeset. `skipIfExtensionNotPresent` relies upon a testEnv.var which is now, by design, not available until the `testEnv` is booted (previously it was relying on an envvar which was there from the start). Previously `getInstalledExrwnsions` was on browser, and `skipIfExtensionNotPresent` was not. One of them always had to be not in `defaultFunctions` so it could reference the browser object with the other method already present on it. I've switched which one is on browser here as `akipIfExtensionNotPresent` also needs access to the `testEnv` which is within scope in `defaultFunctions`. Hope that makes sense. It is confusing, though the code here is I think no less clear than what was there before, and this change is necessary in this branch.

test/setup/loadLocalDockerImage.ts
lorenjohnson: Currently all of our Docker-related spawns are sync (e.g. `testEnv.up`, `down, etc), so it's in line with that.  I for now prefer that, but could  be easily convinced to use the async child_process `spawn` instead of `spawnSync` if I were also going to be changing all the other Docker-related spawns to async. Let me know if you have reason to think it should be async. 

For now using blocking sync operations for booting up the Docker / test environment stuff feels best to me, but marginally. I could have went either way. For now though I wouldn't be excited to factor each of those places to async methods, and then re-test everything. If we find a reason we don't like it later it wouldn't be much to amend though.

docs/topics/testing.md
```diff
@@ -7,53 +7,51 @@ You can run the tests in the docker container locally. Testing locally requires
...
+./test.sh repo
 ```
 
 In order to test your own instances of the services, make sure to set the following environment variables to the services that should be tested. 
 
 
 ## To only run a specific file in the repo tests
 ```
-make test SUITE=repo FILTER=babel*
+./test.sh repo --spec babel*
```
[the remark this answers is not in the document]
lorenjohnson: Yep. I will do that. 

\* Again though this search string and wildcards will once work again as described here, in addition to file path, once we update our WDIO version. The bug is currently here and on `main`. Looking at the `CHANGELOG` on WDIO there there have actually been a pretty good amount of bug fixes since we picked version `8.18.2` in October. We may want to just update it...

test/setup/loadLocalDockerImage.ts
lorenjohnson: One more note of clarification here: That there are not async things happening in `beforeServices` is understandably confusing. it is setup as an `async` method because its setup as a test hook and all our test hooks are set to async. Meanwhile our test services docker stuff (image load, up/down, start, stop) is all setup to be sync.

test/specs/pingback/pingback.ts
lorenjohnson: Good eye. Yes, this one is complicated. Previous the value for `PINGBACK_BEACON_SERVER` (would now be `PINGBACK_BEACON_URL`) was set globally in `default.env` (which is now `test-services.env`) but only used once in this test. Meanwhile the host which composes that URL is set literally in `suites/pingback/docker-compose.override.yml` as well as in its URL form in `suites/pingback/LocalSettings.php`, which it was before as well. The decision here is to take a step backward until this is sorted more and make it a literal string in all 3 places instead of only in 2 of those places. I think this leaves the "problem" more clear and gives us space to answer to it if/when it comes up in the future. All of the rest of our global URL variables are named according to core Wikibase Suites services and are different than `PINGBACK_BEACON_URL`. 

Hope that make sense. It's a good catch, but I think at this point this is the right way to leave it vs further sorting within this branch. It is in my opinion no more, and maybe slightly less, ideal than before this PR.

test/specs/repo/extensions/babel.ts
lorenjohnson: Good catch. This is now resolved but for some reason Github didn't catch that.

test/setup/TestEnv.ts
```diff
@@ -0,0 +1,137 @@
...
+		const dockerComposeCmd = `${this.baseDockerComposeCmd} ${dockerComposeOptionsCommandAndArgs}`;
+
+		this.testLog.debug( 'Running: ', dockerComposeCmd );
+
+		const result = spawnSync( dockerComposeCmd, {
+			stdio: 'pipe',
+			shell: true,
+			encoding: 'utf-8',
+			env: { ...this.vars, OUTPUT_DIR: this.settings.outputDir, PATH: process.env.PATH }
+		} );
```
RickiJay-WMDE, line 93: Just curious - what's the thinking behind making this `sync` instead of `async`?

RickiJay-WMDE: Has the knock-on effect of making `dbQuery` `sync`, and I'm wondering what the advantages or disadvantages of that is

test/specs/repo_client/interwiki-links.ts
```diff
@@ -12,28 +12,28 @@ describe( 'Interwiki links', function () {
...
+			pass: testEnv.vars.DB_PASS,
 			database: 'client_wiki'
 		};
 		const clientLink = repoLinkFile
 			.toString()
 			.replace( /<WIKI_ID>/g, 'my_wiki' )
-			.replace( /<HOSTNAME>/g, process.env.MW_SERVER );
+			.replace( /<HOSTNAME>/g, testEnv.vars.WIKIBASE_URL );
 
 		await browser.dbQuery( clientLink, config );
```
RickiJay-WMDE, line 32: Is now `sync`

test/specs/repo_client/interwiki-links.ts
```diff
@@ -12,28 +12,28 @@ describe( 'Interwiki links', function () {
 		const repoLink = repoLinkFile
 			.toString()
 			.replace( /<WIKI_ID>/g, 'client_wiki' )
-			.replace( /<HOSTNAME>/g, process.env.MW_CLIENT_SERVER );
+			.replace( /<HOSTNAME>/g, testEnv.vars.WIKIBASE_CLIENT_URL );
 		await browser.dbQuery( repoLink );
```
RickiJay-WMDE, line 16: Is now `sync`

test/specs/pingback/pingback.ts
```diff
@@ -2,7 +2,7 @@ import assert from 'assert';
 
 describe( 'Pingback', function () {
 	it( 'Should ping on first page request', async () => {
-		await browser.url( process.env.MW_SERVER + '/wiki/Main_Page' );
+		await browser.url( testEnv.vars.WIKIBASE_URL + '/wiki/Main_Page' );
 
 		// eslint-disable-next-line wdio/no-pause
 		await browser.pause( 5 * 1000 );
```
RickiJay-WMDE, line 8: `dbQuery` below is now `sync`

docs/topics/publishing.md
```diff
@@ -83,8 +83,8 @@ Remove locally created artifacts and use the downloaded ones
 
 ```
 $ make clean
-$ cp artifacts/WORKFLOW_RUN_NUMBER/DockerImages/* artifacts/
-$ make test SUITE=<suite-name>
+$ cp artifacts/WORKFLOW_RUN_NUMBER/BuildArtifacts/* artifacts/
+$ ./test.sh <suite-name>
```
RickiJay-WMDE, line 87: This is the only usage of `BuildArtifacts` I can find, but `DockerImages` is still used in quite a few places - is the splitting deliberate, or should there be more or fewer changes?

test/setup/wdio.conf.ts
```diff
@@ -0,0 +1,142 @@
...
+		// Test Configurations
+		// ===================
+
+		// Level of verbosity: "trace", "debug", "info", "warn", "error", "silent"
+		logLevel: settings.logLevel as Options.WebDriverLogTypes,
+
+		outputDir: settings.outputDir,
+
+		// Default timeout for each waitFor* command.
+		waitforTimeout: settings.waitForTimeout as number,
```
RickiJay-WMDE, line 61: Can we either make `waitForTimeout` non-optional or set a default here?

RickiJay-WMDE (changes requested):

test/setup/wdio.conf.ts
RickiJay-WMDE: Did you find this lint instruction to be needed? VS Code didn't object to me at its removal

test/package.json
```diff
@@ -35,8 +41,6 @@
 		"node": ">=10"
 	},
 	"scripts": {
-		"test:run": "wdio ./suites/$SUITE/$SUITE.conf.ts",
-		"test:run-filter": "wdio ./suites/$SUITE/$SUITE.conf.ts --spec $FILTER",
 		"lint": "npx eslint . --ext .ts",
 		"fix": "npx eslint . --fix --ext .ts"
```
RickiJay-WMDE, line 45: Should include the `.js` file added

RickiJay-WMDE (changes requested):

test/specs/upgrade/pre-upgrade.ts
```diff
@@ -1,7 +1,38 @@
...
+				]
+			};
+
+			const itemId = await WikibaseApi.createItem(
+				getTestString( itemLabel ),
+				data
+			);
+
+			assert.strictEqual( itemId.startsWith( 'Q' ), true );
+			assert.strictEqual( propertyId.startsWith( 'P' ), true );
```
RickiJay-WMDE, line 32: Is there a reason to use `assert.strictEqual( X, true)` over `assert( X )`?

test/setup/wdio.conf.ts
lorenjohnson: It wasn't new, just how it was before. Good to know it's not needed anymore, removing.

test/types/wdio.d.ts
```diff
@@ -10,22 +11,17 @@ declare namespace WebdriverIO {
...
-		 * Execute docker command on container and get output
+		 * Skip test if extension is not installed (present) on the Wikibase server
 		 */
-		dockerExecute: (
-			container: string,
-			command: string,
-			opts?: string,
-			shouldLog?: boolean,
-		) => Promise<unknown>;
+		skipIfExtensionNotPresent: ( test: Context, extension: string ) => Promise<void>;
```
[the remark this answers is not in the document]
lorenjohnson: Moving, and made the browser commands in defaultFunctions follow the same order now.

docs/topics/publishing.md
lorenjohnson: Good catch, and interesting. That should definitely be `DockeImages`, as it is on `main` in this doc. Will update and look around as to why git is preferring this to what is on `main``, I didn't change the file directly but Robert and I were working off the same base branch of his "repro builds" before it was merged down so I'm thinking this got missed in a merge resolution.

lorenjohnson: Ok, I diff'd with main and this is the only thing different in that doc.

test/setup/wdio.conf.ts
lorenjohnson: The default values for this are set in `makeTestSettings` and come from `test-runner.env` which is equally true for `testTimeout`, `logLevel` and a few others in here. `wdio.conf.ts` is a bit of a shell at this point, and on purpose. I've added the comment to that effect from my note in this PR  to the top of this file to make that more clear.  Also have now made this and all other `TestRunnerSettings` properties required (except `runHeaded` which is truly optional.

test/package.json
lorenjohnson: Actually I just made another pass at TS'ing this file and after doing less with `yargs` than I was initially this went well. Now `cli.ts` with a bit of clean-up in there to make it a bit clearer.

test/specs/upgrade/pre-upgrade.ts
lorenjohnson: Probably not, but this is not new code, just moved, so I'm going to let it be for now.

test/setup/TestEnv.ts
lorenjohnson: I think my previous explanation breezed over this a bit... I share your ambivalence, and am open to switching to the `child_process.sync`, but don't prefer it for these reasons:

Reason 1: In `TestEnv` I'm only spawning Docker commands and in all cases we do always want to wait for their completion before executing anything else in the Node side. This stuff is happening exactly at the boundary between our "test runner" / test code and the "test services" we are testing and managing through Docker. `TestEnv`'s job is largely to handle all the things that are not about running the test strictly speaking, but are about setting up the environment in which the tests runs, which is largely a Docker world in our setup. I think you get all of that, but lay it out for context. So using `spawnSync` is helpful here as it ensures that we block Node execution at each of these steps when we otherwise, as a rule, want to make sure we do that. This is probably true with most of our async functions as well, but I don't mind and kinda like it being enforced and highlighted in the Docker / `TestEnv` stuff here exactly because it is a crossing of worlds. 

Reason 2: It is a little unexpected but Node's `child_process.spawn` (vs `spawnSync`) is async but isn't implemented to the async/await pattern. The API is not particular ergonomic and would likely warrant another helper function to re-package `child_process.spawn` in an async/await way, or use one of a few commonly used npm modules that would to that for us. However, pragmatically the effect would be the same in our use. We would alway want to await the Docker command to fully complete before proceeding executing any test code. That said, if Node's async `spawn` function was implemented to the async/await interface already I'd probably just have used that to make it look like everything else.

lorenjohnson: Here is what I've done: I've went ahead and converted `runDockerComposeCmd` to be async while still using `spawnSync` internally. This keeps `browser.dbQuery` async like all the other `browser` commands as well as makes the private methods `startServices` and `stopServices`  on `TestEnv` also async. This should make things more consistent with the rest of our code while not unnecessarily getting into using an async `spawn` setup.

RickiJay-WMDE (approved):

PR #543, opened 2023-12-16
T353520: Make Property spec less flakey applying WDIO Intercept

lorenjohnson:
https://phabricator.wikimedia.org/T353520

Unfortunately using the WDIO Intercept service doesn't on its own yet make the Property spec entirely not flakey. There are a several things going on here: 

1) Not checking if the "save" button is enabled before clicking, which is remedied here by a better selector in `Property.page.ts`
2) Pending jobs from the API calls in the initial test, remedied by a `waitForJobs` call at the end of that test.
3) AJAX requests being issued once a Statement or Reference is added which don't complete before the page is reloaded, which theoretically should be remedied by the WDIO Intercept service waiting for all AJAX to complete before loading the page. 
4) Possible caching--though I think I have mostly eliminated this as a root cause.

Waiting for AJAX calls to complete does seem to make things mostly reliable (and adds 30 seconds to the test run time). Unfortunately though I have seen at least once the added Reference still not appear on the page after reload with Intercept added. 

In the end I think the best scenario here may be to not use the intercept and simply employ a 2 second browser pause in the `afterTest`. This I suspect will be the most reliable and not add the 30 second overhead for waiting out AJAX calls. However, despite that it's not solving all our problems here, I think the WDIO Intercept service is worth being aware of maybe still including in our test setup.

### DIFF
--- a/test/helpers/default-functions.ts
+++ b/test/helpers/default-functions.ts
@@ -315,6 +315,17 @@ export function defaultFunctions(): void {
 			);
 		}
 	);
+
+	browser.addCommand(
+		'waitForPendingRequests',
+		async (): Promise<boolean> => {
+			return browser.waitUntil(
+				async () => {
+					return !!browser.hasPendingRequests();
+				}
+			);
+		}
+	);
 }
 
 export async function skipIfExtensionNotPresent(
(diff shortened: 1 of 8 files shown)


lorenjohnson:
Why the WDIO Intercept service is something we should be aware of, and potentially employ some future tests, the conclusion here is to leave this out and fix the flakey'ness here in https://github.com/wmde/wikibase-release-pipeline/pull/559 using Mocha retries.

PR #544, opened 2023-12-18
T353625 fix: wdqs out of memory / out of fds

rti:
After an update to NixOS 23.11 which ships Docker 24.0.5 at the moment, running WDQS in the test setup broke. This PR fixes it by setting the `ulimit` for file descriptors explicitly.

### DIFF
--- a/test/suites/docker-compose.yml
+++ b/test/suites/docker-compose.yml
@@ -91,6 +91,13 @@ services:
     image: "${WIKIBASE_SUITE_WDQS_IMAGE_URL}"
     restart: always
     command: /runBlazegraph.sh
+    # Set number of files ulimit high enough, otherwise blazegraph will abort with:
+    # library initialization failed - unable to allocate file descriptor table - out of memory
+    # Appeared on Docker 24.0.5, containerd 1.7.9, Linux 6.6.6, NixOS 23.11
+    ulimits:
+      nofile:
+        soft: 32768
+        hard: 32768
     networks:
       default:
         aliases:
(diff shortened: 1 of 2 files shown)


test/suites/docker-compose.yml
```diff
@@ -91,6 +91,13 @@ services:
     image: "${WIKIBASE_SUITE_WDQS_IMAGE_URL}"
     restart: always
     command: /runBlazegraph.sh
+    # Set number of files ulimit high enough, otherwise blazegraph will abort with:
+    # library initialization failed - unable to allocate file descriptor table - out of memory
+    # Appeared on Docker 24.0.5, containerd 1.7.9, Linux 6.6.6, NixOS 23.11
+    ulimits:
```
lorenjohnson, line 97: Only one question: These limits are best done here instead of changing settings and expectations on the running Docker environment? If so it seems like these settings also need to get updated in `example/docker-compose.yml` also/eventually. Thoughts?

rti: yes, apparently this has to be done from outside the container
https://stackoverflow.com/questions/24318543/how-to-set-ulimit-file-descriptor-on-docker-container-the-image-tag-is-phusion

rti: Will add the change to example

rti: done @lorenjohnson

lorenjohnson: Sounds good! And yes that makes sense that they can't be defined by the image. I was curious more about what common default ulimits are on Linux distros, etc and from cursory research looks like 2-8k is common. And I did see in the Blazegraph docs an explanation why they need to be high in Linux environments and the suggestion to up the limits, etc. And finally, thanks for adding to Example. I think that does makes sense👌🏼

lorenjohnson (approved):
🏁

PR #557, opened 2023-12-23
T353520: Property and Special:Property no longer flakey

lorenjohnson:
Uses Mocha "retries" for Property test, and WDIO `browser.waitUntil()` for Special:Property. 

This fixes flakiness by expecting things to be a little latent and continuing to try until expected results appear. I think this the simplest and most clear solution of the options evaluated, as well as has proved to be the most reliable in my testing.  I don't think we'll continue to see failures in this setup. Slightly better than a long `browser.pause()`, which could achieve the same thing for Special:Property, but this way is more descriptive and doesn't always add the full base "pause" time to the tests. In the case of Property we need to make sure the page gets reload to check for the values each time, so `this.retries( 4 )` on the related tests also will reload the page up to 4 times looking for the new value.

### DIFF
--- a/test/cli.ts
+++ b/test/cli.ts
@@ -1,5 +1,6 @@
 /* eslint-disable no-use-before-define */
 import { Launcher, RunCommandArguments } from '@wdio/cli';
+import logger from '@wdio/logger';
 import lodash from 'lodash';
 import chalk from 'chalk';
 import yargs from 'yargs';
@@ -159,10 +160,16 @@ export async function runWdio(
 	wdioOpts: Partial<RunCommandArguments>
 ): Promise<number> {
 	try {
+		// `logger` is a singleton and without this line the `<suiteName>/results/wdio.log` of the
+		// first suite in a multiple suite test ran (e.g. `./test.sh all`) is appended for all the
+		// suites in the run
+		logger.clearLogger();
+
 		const wdio = new Launcher(
 			configFilePath,
 			wdioOpts
 		);
+
 		return wdio.run();
 	} catch ( e ) {
 		throw new Error( 'Failed to start the test suite: ' + e.stacktrace );
--- a/test/specs/repo/property.ts
+++ b/test/specs/repo/property.ts
@@ -43,6 +43,8 @@ describe( 'Property', function () {
 			} );
 
 			it( 'Should be able to see added statement', async () => {
+				this.retries( 4 );
+
 				await $( '=STATEMENT' );
 				const resultStatement = await $(
 					`aria/Property:${stringPropertyId}`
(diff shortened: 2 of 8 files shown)


test/cli.ts
```diff
@@ -159,10 +160,16 @@ export async function runWdio(
 	wdioOpts: Partial<RunCommandArguments>
 ): Promise<number> {
 	try {
+		// `logger` is a singleton and without this line the `<suiteName>/results/wdio.log` of the
+		// first suite in a multiple suite test ran (e.g. `./test.sh all`) is appended for all the
+		// suites in the run
+		logger.clearLogger();
```
lorenjohnson, line 166: ^^ Was very happy to finally find the source of this bug!

test/specs/repo/property.ts
```diff
@@ -58,21 +51,25 @@ describe( 'Property', () => {
...
 				await browser.keys( stringPropertyId.split( '' ) );
 				await $( propertyIdSelector( stringPropertyId ) ).click();
 				await browser.keys( 'REFERENCE'.split( '' ) );
-				await Property.saveStatement.click();
+				await Property.saveStatementLink.click();
 			} );
 
-			it( 'Should be able to see added reference', async () => {
+			it( 'Should be able to see added reference', async function () {
+				this.retries( 4 );
```
lorenjohnson, line 64: Will retry this up to 4 times before calling it a failure. The number of retries, if more than 1, will be reported parenthetically by the Spec reporter at the end.  Note that this runs the `beforeTest` before retrying the test each time as well, so the page is reloaded each time.

test/specs/repo/special-property.ts
```diff
@@ -55,18 +55,20 @@ describe( 'Special:NewProperty', function () {
...
+				limit: 1000
+			} );
+			numberOfPropertiesAfter = await SpecialListProperties.properties.length;
+			return numberOfPropertiesAfter === numberOfPropertiesBefore + 1;
+		}, {
+			// wait for the $wgWBRepoSettings['sharedCacheDuration'] cache to
+			// timeout, so the list of properties reflects the change
+			interval: 1000,
+			timeout: 10000,
+			timeoutMsg: 'expected total number of reflect new property in list within 10 seconds'
```
lorenjohnson, line 71: ^^ Similar to `mocha.retries()` this is utilising the WDIO `waitUntil()` method to do a similar thing but within the test. Note that the page will reload at the top of the loop. The default timeout for `waitUntil` is 5 seconds, and the interval set here is 1 second. So this will try to pass once Immediately, and again at 1, 2, 3, 4, and 5 seconds, before failing.

darthmon (approved):
checked together with Lojo

PR #558, opened 2023-12-27
Moves from waitForUrls to Docker healthchecks in test setup

lorenjohnson:
Hands the reporting of the "Ready" status to docker services back to the build-it docker feature for this purpose.


test/test-services.env
```diff
@@ -1,19 +1,32 @@
 # Any of these can be overridden locally in ../local.env
```
lorenjohnson, line 1: No values are changed below, but reorganised to bring the URL and HOST values near to each other and to the top. Some additional commenting.

test/suites/example/docker-compose.override.yml
```diff
@@ -15,11 +15,38 @@ services:
       - ../../example/extra-install.sh:/extra-install.sh
     environment:
       - QS_PUBLIC_SCHEME_HOST_AND_PORT=http://quickstatements.svc:80
```
lorenjohnson, line 17: NOTE: These override health checks are here so this can be released independently, but go away in a subsequent PR (the nginx proxy/ Quickstatements fix one) which applies them directly in the /example docker compose.

darthmon (approved):
Went through the code together with Loren

PR #559, opened 2023-12-27
T353520: Enables debugging in Node of spec files in test runs

lorenjohnson:
Using `DEBUG=true ./test.sh ...` turns on Node inspector for the spec file runs enabling VS Code Debug console breakpoints. Related `.vscode/launch.json` included for this purpose. The current functionality is not changed by this but the way we run `cli.ts` / `wdio` changes in the way needed to allow enabling inspector when the `DEBUG` flag is present.


test/docker-compose.yml
```diff
@@ -16,12 +16,16 @@ services:
       dockerfile: Dockerfile
     stop_grace_period: 1s
     entrypoint: /bin/bash
+    ports:
+      - "9229:9229"
```
lorenjohnson, line 20: The default Node debug / inspector port.

darthmon (approved):
reviewed together with Loren live

PR #562, opened 2023-12-30
T349806 (3) - Quickstatements OAuth fix with reverse proxy added to Example

lorenjohnson:
⚠️ Branched from #563 and includes that work. This branch can be merge to `main` after both PRs are approved. Related to https://phabricator.wikimedia.org/T349806 and https://phabricator.wikimedia.org/T354073. 

* Adds [nginx-proxy](https://github.com/nginx-proxy/nginx-proxy) to example setup along with a (temporary) top-level domain (TLD) to make all Wikibase Suite (WBS) applications and services accessible via host names common between the Docker network and the host machine. This solves the Quickstatement issues, and makes WBS services addressable on boot at the addresses specified (`template.env` / `.env`) vs `localhost:PORT` (i.e. http://wikibase.berlinresidency.art, http://quickstatements.berlinresidency.art, and http://wdqs-frontend.berlinresidency.art:9999).
* Updates related `example` test suite.
* Further refines the organisation and comments in `example/template.env` and the `example/docker-compose.*` files. 

Remaining TODOs:

- [ ] Integrate `wbs-precheck` into `./wbs` as a `./wbs precheck` command option. Also run `precheck` as a first  install step when running `./wbs` for the first time before creating the `.env` file. When the pre-check doesn't pass in that context don't create `.env`, but prompt with "Checks didn't pass, try and run WBS anyway? (N)" and only create it if responding (Y). Create a Phab ticket for this work. 

- [ ] Start a draft of https://www.mediawiki.org/wiki/Wikibase/Docker which reflects usage of the `./wbs` script and keeps updating as we make any further refinements such that it is ready for review by Dan before release.  * This can be done by simply copying the WikiText into a new draft page on your own user and then sharing that with Dan once we feel it's ready. This is what he has recommended as a workflow in the past.  Create a single Phab ticket for this and the subsequent documentation items. 

- [ ] Document how to resolve pre-check failing the `wikibase.local.gd` subdomain access check. Add relevant comments in `example/.env.template` and documentation not `example/README` and https://www.mediawiki.org/wiki/Wikibase/Docker accordingly. Here is some background and the known solutions as a solid start on the docs about this:
  -  **Background:** *DNS entries which resolve to `127.0.0.1` are rejected in some environments due to [DNS Rebinding](https://en.wikipedia.org/wiki/DNS_rebinding) Protection happening within the operating system or as part of the network router setup. 
  - **Solution 1:** Just use port-based access for each service. These addresses will be printed-out as the URLs for accessing the services by `./wbs` when the `wikibase.local.gd` access precheck fails. Note however that when this is the case QuickStatements won't be able to authorise, and therefore not useful. To make it also available proceed with one of the subsequent solutions.
  - **Solution 2:** A reliable work around is to use Firefox and manually select to use CloudFlare DNS within the browser Security settings.
  - **Solution 3:** Elect to use a Public DNS server instead of those provided by the Internet service providers automatic configuration on your computer. CloudFlare and Google provide DNS servers which are commonly used for this purpose. Choosing either of these should clear-up any issue for the use of `wikibase.local.gd`, etc. However, this implicitly does mean you're removing a layer of protection so it's a user choice.*

- [ ] Decide what notes we make, if any, about running WBS Example on Apple Silicon-based Macs. If we want to make a statement about this now, then craft that statement both in https://www.mediawiki.org/wiki/Wikibase/Docker and in example/README.


example/docker-compose.extra.yml
```diff
@@ -62,25 +69,26 @@ services:
...
         aliases:
-         - wdqs.svc
+         - ${WDQS_HOST}
     environment:
+      - VIRTUAL_HOST=${WDQS_HOST}
+      - VIRTUAL_PORT=${WDQS_PORT}
       - WIKIBASE_HOST
-      - WDQS_HOST=wdqs.svc
-      - WDQS_PORT=9999
-    expose:
```
lorenjohnson, line 70: Note: The `expose` key on a Docker Compose service is mostly just a notation, and has no practical effect in our setup. The port 9999 is configured for the WDQS service, and available on that container regardless of the `expose` value. All ports on a container are by default available to other nodes in the Docker network.

example/template.env
```diff
@@ -1,55 +1,62 @@
-## Example / Template .env file for Wikibase release pipeline docker compose example
+# Example / Template .env file for Wikibase release pipeline docker compose example
 # WARNING: Do not add comments on the same line as env vars, as in some environments they will be included in the var!
 
-## Image Configuration
```
lorenjohnson, line 4: These are moved to the bottom of the file as they are of top-line interest to us in updating this template, but of bottom-line interest to the user of this file.

example/docker-compose.extra.yml
lorenjohnson: Yes, it is always just a documentation feature. We were not using the key for any of the other services (Elasticsearch, WDQS proxy, etc) so think it's better to take it out or notate it on all the relevant services for consistency. I think we should just leave it out as I think its misleading and becomes the only param in our Docker compose setup that has no purpose other than documentation. 

This from ChatGPT is helpful context as well, I suspect this related to why it was there I the first place:

> Me: Was "expose" used in earlier versions of Docker compose in a different way such that it was more than documentation and had additional practice effects?
> 
> ChatGPT: In earlier versions of Docker Compose, specifically in versions 1 and 2, the expose key had a more significant impact compared to its role in version 3. In Docker Compose version 1 and 2, the expose key not only documented the ports but also implicitly mapped them to the host machine. This meant that if you used expose without specifying the ports key, the service's ports would be exposed on the host machine as well. 

Because of this you'll notice if you Google on this topic there is a lot of conflicting and I think often misleading information. For all these reasons I think we just don't "expose".

PR #566, opened 2024-01-23
T33704 - Page Object Updates

RickiJay-WMDE:
- Renamed page objects to include `Page`
- Added JSDocs on each page `open()` method to specify url and any parameter defaults
- Wrote `urlParameters` method for transforming Objects into key-value pairs formatted for urls
- Swapped `browser.url` for `*Page.open` wherever possible - we had unnecessary `browser.url` calls when pages that already existed fit the bill
- Removed single-use or no-use page getters
- Implemented a `SpecialListPropertiesPage.open`
- `ItemPage` for `/wiki/Item:{id}` - we use this everywhere, it seemed like a no-brainer
- The Not-Page Page: Wrote `SpecialEntityDataPage` to handle data requests to `/wiki/Special:EntityData`, which are fairly widely used


lorenjohnson (approved):
👍🏼 Looks good as per what is described in the ticket. I do think there is a conversation about test clarity and the use of Page objects to still be had, but can wait until another time.

PR #567, opened 2024-01-24
T349806 - Update to Example config and docs, adds WBS script

lorenjohnson:
Work relates to https://phabricator.wikimedia.org/T349806 and https://phabricator.wikimedia.org/T354073. 

**CHANGELOG**

* Adds Docker Healthcheck to compose files
* Aligns `*_IMAGE_NAME ` Docker image name variables to the new `WIKIBASE_SUITE_*_IMAGE_URL` standard used in our build and release environment.
* Adds `example/.env.defaults` for less used settings which can be "upgraded" within source control (in particular versioned `WIKIBASE_*_IMAGE_URL`s)
* Refines the organisation and comments in `example/docker-compose.*` files
* Adds [nginx-proxy](https://github.com/nginx-proxy/nginx-proxy) to example setup along with an optional top-level domain (TLD) to make all Wikibase Suite (WBS) applications and services accessible via host names common between the Docker network and the host machine. This solves the QuickStatement issues, and makes WBS services addressable on boot at the addresses specified (`template.env` / `.env`) vs `localhost:PORT` (i.e. http://wikibase.local.gd, http://quickstatements.local.gd, and http://wdqs-frontend.local.gd:9999).
* Adds optional `example/wbs.sh` utility script for consistently starting, stopping, and resetting the WBS example installation and configuration
* Updates `README`
* Updates related `example` test suite.

**Remaining TODOs**

- [ ] Confirm that configuring Docker network aliases which are the same as externally routed addresses works in a HTTPS production setup, if doesn't add to comments and docs a note about how to handle that in production
- [ ] Decide whether to include `wbs-precheck` and `wbs` with these improvements, and if so how to frame them
- [ ] Review, copy edit, and update README, then tag Dan for final review/edits
- [ ] Consider making this README and the Docker Install page the same, and regardless, create a new draft of the Docker Install page in WikiText to reflect any changes in this PR, once done send that to Dan for final review/edits
- [ ] Test VPS needs a more recent version of Docker Compose. Follow instructions at: https://docs.docker.com/engine/install/debian/#install-using-the-repository


example/docker-compose.extra.yml
```diff
@@ -120,9 +125,9 @@ services:
         aliases:
          - ${QUICKSTATEMENTS_HOST}
```
lorenjohnson, line 126: It is subtle, and worth noting here and perhaps somewhere in our docs. How this magic works is that we are making the `QUICKSTATEMENTS_HOST` (e.g. "quickstatements.local.gd") a Docker network alias for this service. This means that when inside the docker container when the Wikibase service goes to "http://quickstatements.local.gd" it will go to this container. Conversely on the host, when the nginx-proxy is running, it will route to this container because of the `VIRTUAL_HOST` being set to the same "quickstatements.local.gd" address.

example/docker-compose.extra.yml
```diff
@@ -62,28 +73,26 @@ services:
...
       default:
         aliases:
-         - wdqs.svc
+         - ${WDQS_HOST}
     environment:
       - WIKIBASE_HOST
-      - WDQS_HOST=wdqs.svc
-      - WDQS_PORT=9999
-    expose:
-      - 9999
```
lorenjohnson, line 71: `expose` has not functional purpose in Docker compose files outside of documentation. All ports on a docker service are by default exposed to all other services in the same network. This was different in 1.x versions of Docker and Docker compose, and likely why this setting was a hanging around.

PR #574, opened 2024-02-06
T350878 - Uncovered Tests

RickiJay-WMDE:


test/specs/fedprops/prefetching.ts
```diff
@@ -85,4 +85,60 @@ describe( 'Property Prefetching', function () {
 		// +1 for ?
 		assert.strictEqual( ( await $$( 'ul.special li' ) ).length, NUM_PROPERTIES + 3 );
 	} );
+
+	it( 'Should be able to change limit', async () => {
+		await browser.url(
+			`${testEnv.vars.WIKIBASE_URL}/wiki/Special:RecentChanges?limit=50&days=7&urlversion=2&enhanced=0`
+		);
```
lorenjohnson, line 92: The thing I have in mind is that we make the base Page object look like this (maybe renaming to `MediaWikiPage` or `WikibasePage`):
```
export class Page {
	public async open( path: string ): Promise<void> {
		await browser.url( `${browser.options.baseUrl}/${path}` );
	}
}

export default new Page();
```
And then do:
```
Page.open( 'Special:RecentChanges?limit=50&days=7&urlversion=2&enhanced=0' )
```
In all the cases where we a unique Page object is not justified. 

I have a couple other thoughts in line with that more global chat about Page Object we touch-on the other day. I'm happy to get into further with you anytime you feel find the right moment for it 👌🏼

test/specs/repo/property.ts
```diff
@@ -92,6 +93,63 @@ describe( 'Property', function () {
...
+				await browser.keys( `${dataType.name} Description`.split( '' ) );
+				await $( 'label=Aliases:' ).click();
+				await browser.keys(
+					`${dataType.name} Alias A|${dataType.name} Alias B`.split( '' )
+				);
+
+				await $( 'span=Set label, description and aliases' ).click();
+			} );
+
+			it( 'Should be able to see updated label, description, aliases', async () => {
```
rti, line 139: combine with above test?

test/specs/repo/queryservice.ts
```diff
@@ -151,4 +152,75 @@ describe( 'QueryService', () => {
...
+		);
+
+		await QueryServiceUIPage.open( `SELECT (COUNT(*) AS ?count)
+		WHERE {
+		  ?item <${testEnv.vars.WIKIBASE_HOST}/prop/direct/${propertyId}> <${testEnv.vars.WIKIBASE_HOST}/entity/${itemId}> .
+		}` );
+
+		expect(
+			( await QueryServiceUIPage.resultTable.$( 'tbody' ).$$( 'tr' ) ).length
+		).toBeGreaterThan( 0 );
```
rti, line 224: Maybe we can check a bit more here? Like the actual values in the table?

rti (changes requested):
Looking good, just some minor comments. Thanks a lot for your work here. 🎉🎉🎉

test/specs/fedprops/prefetching.ts
```diff
@@ -85,4 +85,48 @@ describe( 'Property Prefetching', function () {
...
+			$( 'div.mw-rcfilters-ui-changesLimitAndDateButtonWidget' )
+		).toHaveTextContaining( '7 days' );
+		await $( 'div.mw-rcfilters-ui-changesLimitAndDateButtonWidget' ).click();
+		await $( 'div.mw-rcfilters-ui-datePopupWidget-hours' ).$( 'span=2' ).click();
+		await expect(
+			$( 'div.mw-rcfilters-ui-changesLimitAndDateButtonWidget' )
+		).toHaveTextContaining( '2 hours' );
+	} );
+
+	it( 'Should be able to change time to 3 days', async () => {
```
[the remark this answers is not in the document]
RickiJay-WMDE: moved to repo/recent-changes.ts

test/specs/repo/property.ts
RickiJay-WMDE: Combined!

test/specs/repo/queryservice.ts
RickiJay-WMDE: Updated!

rti (approved):

lorenjohnson (approved):
Hey @RickiJay-WMDE this looks good 👍🏼 Two questions which I'd love to loop in what you on, but certainly neither needs to hold-up marging:

1) Did you follow what I was saying here, and do you have any thoughts: https://github.com/wmde/wikibase-release-pipeline/pull/574#discussion_r1481628281

2) The refactoring of the Elasticsearch tests looks great, happy we're settling on the expects vs the asserts. However it wasn't apparent to me what changed materially in the test, and if you get a minute would like to understand better.

test/specs/fedprops/prefetching.ts
RickiJay-WMDE: If I understand correctly, that's basically what `Page` already does, and we could fully use `page.Open('Special:RecentChanges')` now. That, and a few other things, I'd like to implement in a story to clean up this code some more.

RickiJay-WMDE:
>     2. The refactoring of the Elasticsearch tests looks great, happy we're settling on the expects vs the asserts. However it wasn't apparent to me what changed materially in the test, and if you get a minute would like to understand better.

The explanation for that refactored test:
The search term `Test` returned first `Q36`, which was labeled `test-image-[etc]`, then both `Q36` and `Q37`. We were trying for just `Q37`, which was labeled `testImage-[etc]`. So I changed the search term to `Testimage`, which would only match `Q37`, and double-checked that the search term was appropriate.
I have no idea how this worked previously.

PR #580, opened 2024-02-15
streamline mediawiki version number variables

rti:
Our `variables.env` file contained the mediawiki version in several variables. This PR condenses this into a single variable which is enough to archive the same functionality.


lorenjohnson (approved):
Excellent! Thank you, and LGTM ✨

PR #586, opened 2024-02-21
T358307 - Update Release Variables Script

RickiJay-WMDE:
Updates the commit hashes in `variables.env`, specifically from Gerrit, GitHub, and Bitbucket. The first, it parses the relevant webpage using BeautifulSoup; the latter two, it pulls the relevant data from their respective APIs.

While I did need to tweak some of the urls provided for consistency (and accuracy, in the case of WDQS GUI - the link was broken), I did not commit `variables.env` updated hashes, so that devs can test the script.

Notes:
* Mediawiki extensions include the compatible version of Mediawiki in the URL. I have the script fetch this from the `MEDIAWIKI_VERSION` variable and automatically update the commented URLs to the correct version.
* As mentioned, for both GitHub and Bitbucket, the script pulls data from their respective APIs. The base URLs for each link are changed to `https://api.github.com/repos/` and `https://bitbucket.org/!api/2.0/repositories/` respectively. I didn't make this change in `variables.env`, because I assumed we wanted that file to link to human-readable pages rather than API output.


lorenjohnson (changes requested):
Looks great, with one caveat. Yes, effectively currently our only dev dependency is Docker. Probably should stick with that unless there is compelling reason not to. See here for how we currently run your other Python script, and apply that here?:

https://github.com/wmde/wikibase-release-pipeline/blob/f8a4ea71b0403791230889b259146a164f7dc607/lint.sh#L43-L45 

Perhaps add a "update_hashes" task to the build.sh script to run the Python script in the test (or build) container?

PR #588, opened 2024-02-26
Cleanup II

RickiJay-WMDE:
NOTE: DELETE `test/node_modules/` DIRECTORY AND DOCKER IMAGE `wikibase-suite-test-runner-test-runner`.

So, to be clear, >95% of this - by lines - is formatting and linting cleanup.

#### Formatting & linting changes:

- Formatting `.ts`, `.mjs`, `.json` files - had to add to a few overrides to format JSONs correctly
- Updated all `eslint` packages and plugins
- Somewhere along the way they added a rule about space inside curly brackets
- Added `prettier` and `prettier-plugin-organize-imports` - make formatting consistent, organize imports
- Added `yaml-eslint-parser` - formatting and linting YAML files with `prettier` and `eslint`

#### Actual code changes:

- Every test that was `async () => {}` is now `async function () {}`. Ran into an issue that this fixed a few weeks ago - the context is different between the two, so things like `this.retries()` doesn't actually do anything.
- Speaking of which, added `retries` to the flaky `special-property.ts` test.
- The `lint.sh` script was wrong - it was trying to lint twice, because it previously linted both `/test` and `/docs`
- Updated `engine` to `14.14` - `eslint` caught that we were using features too advanced for `10`, specifically `fsPromises`
- Added `lint:fix:lint`, which just runs `eslint`, and `lint:fix:format`, which just runs `prettier`, and changed `lint:fix` to run both


rti (dismissed):

rti (commented):
Is this done by a linting script?

.github/workflows/_build_test.yml
```diff
@@ -152,3 +151,12 @@ jobs:
...
           path: test/suites/**/results
+
+  test-success:
+    runs-on: ubuntu-latest
+    needs:
+      - test
+      - test-upgrade
+    steps:
+      - run: |
+          echo "hello world"
```
rti, line 162: maybe we can have some explanation here why we add this noop step

rti, line 162:
```suggestion
      - run: true
```

rti (changes requested):

.github/workflows/_build_test.yml
```diff
@@ -152,3 +150,12 @@ jobs:
         with:
           name: TestResults
           path: test/suites/**/results
+
+  test-success:
+    # Post-test CI step - to check all tests succeeded
```
rti, line 155:
```suggestion
    # Post-test CI step - to check all tests succeeded
    # Github Branch Protection rules require this to pass to allow merging
```

RickiJay-WMDE:
> Is this done by a linting script?

Yeah - `prettier`, followed by `eslint`. Having trouble getting the latter to work on yml files, so I haven't added those to the script

RickiJay-WMDE:
> > Is this done by a linting script?
> 
> Yeah - `prettier`, followed by `eslint`. Having trouble getting the latter to work on yml files, so I haven't added those to the script in `package.json`

Update: Got YAML files working

rti (commented):
Shouldn't prettier be installed automatically?

```
❱ ./lint.sh -f             
Fixing Linting and Formatting issues in Typescript
[+] Building 66.4s (15/15) FINISHED
=> naming to docker.io/library/wikibase-suite-test-runner-test-runner
Need to install the following packages:
prettier@3.2.5
Ok to proceed? (y) 
```

rti:
Question: do we lint from the project root now? Or is this another topic?

RickiJay-WMDE:
> Question: do we lint from the project root now? Or is this another topic?

I think that's another topic - we're still linting from `/test` here

rti (approved):

PR #598, opened 2024-03-02
Expect

RickiJay-WMDE:
Replaces `assert` with `expect` for consistency's sake

### DIFF
--- a/test/specs/elasticsearch/elasticsearch.ts
+++ b/test/specs/elasticsearch/elasticsearch.ts
@@ -37,9 +37,9 @@ describe( 'ElasticSearch', function () {
 		let searchResult: SearchResult[];
 
 		const testLabel = 'Testitem';
-		expect( itemLabel ).toEqual( expect.not.stringContaining( testLabel ) );
-		expect( itemLabel.toLowerCase() ).toEqual(
-			expect.stringContaining( testLabel.toLowerCase() )
+		expect( itemLabel.includes( testLabel ) ).toBe( false );
+		expect( itemLabel.toLowerCase().includes( testLabel.toLowerCase() ) ).toBe(
+			true
 		);
 
 		await browser.waitUntil(
--- a/test/specs/elasticsearch/elasticsearch.ts
+++ b/test/specs/elasticsearch/elasticsearch.ts
@@ -37,9 +37,9 @@ describe( 'ElasticSearch', function () {
 		let searchResult: SearchResult[];
 
 		const testLabel = 'Testitem';
-		expect( itemLabel.includes( testLabel ) ).toBe( false );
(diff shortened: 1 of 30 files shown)


test/specs/elasticsearch/elasticsearch.ts
```diff
@@ -37,9 +37,9 @@ describe( 'ElasticSearch', function () {
 		let searchResult: SearchResult[];
 
 		const testLabel = 'Testitem';
-		expect( itemLabel.includes( testLabel ) ).toBe( false );
-		expect( itemLabel.toLowerCase().includes( testLabel.toLowerCase() ) ).toBe(
-			true
+		expect( itemLabel ).toEqual( expect.not.stringContaining( testLabel ) );
+		expect( itemLabel.toLowerCase() ).toEqual(
+			expect.stringContaining( testLabel.toLowerCase() )
```
rti, line 42: I think those nested `expect`s are not really intuitive. Is there a more clear way to express that?

test/specs/upgrade/queryservice-post-upgrade.ts
```diff
@@ -93,14 +92,14 @@ describe( 'Wikibase post upgrade', function () {
...
-		assert( siteLinks !== null );
-		assert( identifiers !== null );
-		assert( timestamp !== null );
-		assert( statement !== null );
+		expect( dateModified ).toEqual( expect.anything() );
+		expect( schemaVersion ).toEqual( expect.anything() );
+		expect( siteLinks ).toEqual( expect.anything() );
+		expect( identifiers ).toEqual( expect.anything() );
+		expect( timestamp ).toEqual( expect.anything() );
+		expect( statement ).toEqual( expect.anything() );
```
rti, line 100: which expect are we using? I cannot find `anything()` in [wdio doc](https://webdriver.io/docs/api/expect-webdriverio) but in [jest doc](https://jestjs.io/docs/expect)

rti (changes requested):
great, really getting into shape this thing, thanks for your work

test/specs/upgrade/queryservice-post-upgrade.ts
RickiJay-WMDE: You get Jest's matchers and testers for free with wdio's expect: https://webdriver.io/docs/api/expect-webdriverio/#default-matchers

test/specs/elasticsearch/elasticsearch.ts
RickiJay-WMDE: The other way to do it would be `expect( actual.includes( expected ) ).toBe( true )` - would we prefer that?

RickiJay-WMDE: ...actually, yeah, I think that is better

rti (approved):

PR #625, opened 2024-03-27
T355893 allowlist

RickiJay-WMDE:


test/specs/repo/queryservice.ts
```diff
@@ -275,4 +275,52 @@ describe( 'QueryService', function () {
...
+		SELECT * WHERE {
+			service <https://query.wikidata.org/sparql> {
+			  ?wd wikidata_wdt:P31 wikidata_wd:Q976981 .
+			  ?wd wikidata_wdt:P31 ?type .
+			  optional { ?wd rdfs:label ?label. filter(lang(?label) = "en") }
+			  optional { ?type rdfs:label ?typelabel. filter(lang(?typelabel) = "en") }
+		   }
+		
+		}
+		LIMIT 50
```
physikerwelt, line 293: To reduce the load for the sparql endpoint, one could reduce the limit further.

build/WDQS/README.md
```diff
@@ -41,7 +41,7 @@ Note on `BLAZEGRAPH_EXTRA_OPTS`: These are options that are directly passed to t
 
 | File | Description |
 | --- | --- |
-| `/wdqs/whitelist.txt` | SPARQL endpoints that are allowed for federation |
+| `/wdqs/allowlist.txt` | SPARQL endpoints that are allowed for federation |
```
rti, line 44: Not really related to this issue, but I think we should find a way how to take notes about breaking changes in order to be ready to put them in our changelog on release. 

Would the `CHANGES.md` be a good place for that for now? Should we add this change in there already? I created a task so we do not forget about it https://phabricator.wikimedia.org/T361245 . If you think the `CHANGES.md` approach is good for now, we can move this ticket to done right away :)

physikerwelt: Yes, however, in this case, it is a `fixing` change;-)

GreenReaper: So what happens for people who have a whitelist.txt in their setup and upgrade to this version? Will this break their setups (or continue to have them broken if they updated recently)?

physikerwelt: It depends on the environment variable ALLOWLIST. https://gerrit.wikimedia.org/r/c/wikidata/query/deploy/+/952275/1/runBlazegraph.sh
If this environment variable wasn't changed to whitelist. Which was copied by the line changed in this commit it should fix things.

lorenjohnson: If the user has specifically overridden that env var I think the setting they are using will work as is. 

There is one edge case which is if this setting was overwritten to mirror to previous default of "whitelist.txt" without a volume mapping to a local copy of ones own "whitelist.txt" into the container, then it would break as that file would no longer exist in the container file system. However, I think that is a pretty far edge case, of which the solution would be to simple remove the overridden setting and rely on the default.

lorenjohnson (approved):
LGTM!

PR #636, opened 2024-04-03
T361227 - Release WMDE.19

RickiJay-WMDE:


RickiJay-WMDE:
> The commit hashes updated here are they derived via the automatic script, or manually? I think for the patch releases we have to be careful about using the automatic script as latest / HEAD versions of each of the extensions, etc aren't necessarily compatible with previous major MediaWiki releases. It probably was, but confirming that this was considered here?

So all the WMF extensions are branched by MW version - we're only pulling updates from their 1.40 branches.
WDQS Frontend isn't, but in my experience so far it doesn't built when it isn't compatible with the MW version. I've been working on checking that and so far it holds.

lorenjohnson (approved):
LGTM 👍🏼

rti (approved):

PR #637, opened 2024-04-04
Add Release 1.39.6 to `main` Change log

RickiJay-WMDE:


lorenjohnson (approved):
LGTM!  
This is cherrypicked from the WMDE15 release, right?  Also-- while basically immaterial, in principle I think this should also make it into the 3 upcoming releases.

PR #648, opened 2024-04-15
Cleanup 1.40 test fragility

RickiJay-WMDE:
Using `() => {}` instead of `function () {}` for tests means the context is incorrect, so things like `this.retries()` don't work.

We made this change to `main` a while ago - it never got down to mw-1.40


rti (approved):

lorenjohnson (approved):
Was surprised that there was still a place where we were using `.retries()` that was using an arrow function. We had already caught this once. Do you think it was a regression?

PR #652, opened 2024-04-16
1 -- T354269: Move extra files in Example into images

RickiJay-WMDE:


lorenjohnson:
Also some clean-up to Example healthcheck additions: 

1) Standardises formatting and use of `localhost` in healthchecks in test docker-compose instances, and removes redundant Example test suite healthchecks (were there until healthchecks were actually added to Example).

2) Forces Example test suite to run against current builds vs versions specified in `template.env`

lorenjohnson:
@rti @RickiJay-WMDE @adeeri  This is now ready to review / merge. Issue was with the bash error handing in the new `default-extra-install.sh`. The way it was when `MW_ELASTIC_HOST`, etc wasn't defined the script would error and exit early, causing not only the remaining QuickStatements install from not running, but bubbling up to cause the rest of `extra-install.sh` to not run. I have fixed that using the "default value" format in the checks (e.g. `${MW_ELASTIC_HOST:-}` noting the `:-`)

Additionally I've went ahead and for now resolved the issue with the Example test suite by setting the suite to run against the current builds vs those specified in `example/template.env`. I think this is the right call for now, but something we may wish to revisit later to make sure we are in some way confirming the correct settings for the published images in the `template.env`. As it is this will make it so we no longer have to do the after release Example update.

RickiJay-WMDE (commented):
LGTM

lorenjohnson (approved):
@RickiJay-WMDE reviewed my additional commits, I reviewed @RickiJay-WMDE work. This is now approved :)

PR #660, opened 2024-04-22
T363103: Example Pre-flight Check script

lorenjohnson:


example/wbs-precheck
```diff
@@ -0,0 +1,96 @@
+#!/bin/bash
```
rti, line 1: I guess one of the files is there due to renaming hick ups, lets remove one of them. As we are not planning to install them in the system (do we?), I'd go with just `./precheck`. But `./wbs-precheck` would be fine with me too.

example/precheck
```diff
@@ -0,0 +1,96 @@
...
+
+# Store PASS/FAIL results
+RESULTS=""
+
+# Check Docker version
+check_version "Docker" "$MIN_DOCKER_VERSION" "docker"
+check_version "Docker Compose" "$MIN_DOCKER_COMPOSE_VERSION" "docker compose"
+
+# Check domain resolution
+# check_domain_resolution "$WIKIBASE_HOST"
```
rti, line 78: Why is this cool feature disabled? :)

example/precheck
```diff
@@ -0,0 +1,96 @@
...
+
+# Check architecture
+if ! [ "$(uname -m)" == "x86_64" ] || [ "$(uname -m)" == "amd64" ]; then
```
rti, line 81: This check could be simplified. If you want to prepare for arm, maybe lets have the new condition as a comment?

rti (changes requested):
Cool thing.

lorenjohnson:
Hey @rti, this is here as I was just gutting out remaining pieces of previous work which we've not yet covered in our Example work all together. It is ready for review, and your comments noted, but FYI it also isn't something I'll get back to for a minute as its down the priority list a bit... unless we decided to bump it up a bit, but at this moment I am not feeling strongly about that either way.

example/wbs-precheck
RickiJay-WMDE: @rti "hiccups" or antiquated "hiccoughs"

example/precheck
lorenjohnson: Because on initial check this likely won't pass unless we put in our run instructions to first sort-out all your DNS stuff, which I think is unnecessary though not necessarily wrong. This was originally part of using the `*.local.gd` host names as default and was confirming that localhost access was working. That didn't work out so I left this out. I think the move here is to actually remove the commented code for now, and bring is such a feature maybe later. It I think would be distracting at this stage.

example/precheck
lorenjohnson: 1) How would you like it simplified? 

2) I think leaving out any ARM related code until we make an ARM-related update is fine with me here, in fact I would imagine doing away with the architecture check in this script once we support both architectures--but maybe I missed your point on that. 

Say more.

example/wbs-precheck
lorenjohnson: Oh yes, that was entirely unconscious and that was the reason. This was `wbs-precheck` which I prefer as I have reasons to believe we should lean into this `wbs` acronym as a possible identity / rename for `example`. It's a same-paging discussion though so I was renaming to simply `precheck` to slim down the things to discuss here, but shot myself in the foot :) I would prefer to leave it `precheck` for now so will remove the extra file.

example/precheck
```diff
@@ -0,0 +1,81 @@
...
+
+# Store PASS/FAIL results
+RESULTS=""
+
+# Check Docker version
+check_version "Docker" "$MIN_DOCKER_VERSION" "docker"
+check_version "Docker Compose" "$MIN_DOCKER_COMPOSE_VERSION" "docker compose"
+
+# Check architecture
+if ! [ "$(uname -m)" == "x86_64" ] || [ "$(uname -m)" == "amd64" ]; then
```
rti, line 66:
```suggestion
if ! [ "$(uname -m)" == "x86_64" ]; then
```

example/precheck
rti: Totally agree to not prepare future features in comments, it would be distracting.

Find my actual simplification suggestion below

example/precheck
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
```
rti, line 1:
```suggestion
#!/usr/bin/env bash
```
for systems having bash in different locations (e.g. nixos)

example/precheck
```diff
@@ -0,0 +1,81 @@
...
+
+# Minimum versions
+MIN_DOCKER_VERSION="22.0"
```
rti, line 4: Do you remember why we require docker 22? I am on 20 and it works™

rti (changes requested):

example/precheck
rti: Never mind. The `flake.nix` still installed 20.10 because I never updated it. 20.10 is end of life, so not going down to that version sounds reasonable.

rti: BTW, I was running docker daemon 24 with docker client 20, the script only considered the client version. Might be good enough for now, but wanted to flag it.

PR #665, opened 2024-04-24
T348742-better-localsettings-handling

lorenjohnson:


build/Wikibase/entrypoint.sh
```diff
@@ -13,47 +13,47 @@ done
...
-    fi
-
-    # Run extrascripts on first run
-    if [ -f /extra-install.sh ]; then
-        # shellcheck disable=SC1091
-        source /extra-install.sh
-    fi
+# Run MediaWiki install script
+rm -f /var/www/html/LocalSettings.php
+rm -f /var/www/html/LocalSettings.generated/LocalSettings.ph
```
rti, line 24:
```suggestion
rm -f /var/www/html/LocalSettings.generated/LocalSettings.php
```

build/Wikibase/entrypoint.sh
```diff
@@ -13,47 +13,47 @@ done
...
+    --dbuser "$DB_USER" \
+    --dbpass "$DB_PASS" \
+    --dbname "$DB_NAME" \
+    --dbserver "$DB_SERVER" \
+    --lang "$MW_SITE_LANG" \
+    --pass "$MW_ADMIN_PASS" \
+    "$MW_SITE_NAME" \
+    "$MW_ADMIN_NAME"
+
+mv -f /var/www/html/LocalSettings.generated/LocalSettings.php /var/www/html/LocalSettings.generated.php
```
rti, line 37:
```suggestion
mv -f /var/www/html/LocalSettings.generated/LocalSettings.php /var/www/html/LocalSettings.generated.php
rm -rf /var/www/html/LocalSettings.generated/
```

lorenjohnson:
@rti looping back on this "install every time" once again. A few things add up to me thinking we actually only want to generate LocalSettings.php (hence run the install) once as it was before: 

* I found all the needed MediaWiki maintenance scripts needed to update the Admin password and email without necessarily creating a new Admin user... However, if we set these things to the corresponding values in the provided `MW_ADMIN_*` env vars, then we're overwriting any changes made through the MediaWiki UI of these values on each container re-create . That isn't going to do.

* The `DB_*` values are dynamically configured due to the "run install every time", but the values are still effectively static in the Maria DB container. 

* All the remaining MediaWIki settings we currently accept as env vars to the WIkibase service end-up statically generated into the `LocalSettings.php`.

So perhaps if we only run install when there is not already a `LocalSettings.php` in the container we're doing the right thing after all. This means that on a container remove / create `run.php install` will still be ran again, but not on container stop / starts it will not. However, if when the container recreates it always installs we still have the issue with perhaps unwanted updates from `MW_ADMIN_*` values which had since been updated directly in the UI.

Not sure what the solution is yet, but would be good to dig back into this for another round on Monday.  

Initial thought: 

We might want to prefix all "one time" env vars on the Wikibase image with `SETUP_*`, and create and document a process for resetting from those provided values. All values that we genuinely want to configure at the service level ongoing (e.g. `MAX_JOBS` or `WG_JOB_RUN_RATE`, etc) we should make clear by not prefixing them with `SETUP_*`.

lorenjohnson:
Current notes:

## WBS Config Overview

- Adding a shared volume and booting the container will result in having some / all `LocalSettings.*.php` files appear in the mapped wikibase-config shared volume folder

- The only things in `wbs/template.env` will be:
  - `SETUP_*` values which are used in install.php process but not otherwise
  - `*_PORT` values for front-end services
  - `*_PUBLIC_HOST` values for front-end services
  - `*_IMAGE_URL` values for Docker images

## Initial Setup, Reset, and Upgrade pseudodocs

**Initial Setup**

1) Set all `SETUP_*` values in `.env` appropriately
2) `docker compose up --wait`

**Reset Setup**

1) Remove all files in `./wikibase-config`
2) Set all `SETUP_*` values in `.env` appropriately
3) `docker compose up --wait`

**MediaWiki Minor version upgrade**

Happens automatically based upon major-tag `*_IMAGE_URL` specification (e.g. `WIKIBASE_WIKIBASE_IMAGE_URL=wikibase/wikibase:v2).

1) `docker compose down`
2) `docker compose up --wait`

**MediaWiki Major Version Upgrade**

This current process is manual and more or less the same as doing a manual MediaWiki upgrade outside of the WBS provisions. Additionally these instructions imply the site being down while `wikibase-config` differences are reconciled. There is a lot of need/room for improvement here.

1) Change `WIKIBASE_BUNDLE_IMAGE_URL` to specify the upgraded image tag, e.g. `wikibase/wikibase:v2` to `wikibase/wikibase:v3`
2) Review Upgrade Notes or Migration Guide from the specific version you're on to the new version you're going to:

   If there are `wikibase-config` changes then:

   1) Backup contents of the `./wikibase-config` volume and then delete everything from it
   2) Set all `SETUP_*` values in `.env` appropriately (taking care with values which may have already been reset in UI)
   3) `docker compose up --wait`
   4) Manually compare/diff and reconcile differences in files in `/wikibase-config` with backed-up files
   5) `docker compose down` / `docker compose up`

   If there are not `wikibase-config` changes then:

   6) `docker compose down` / `docker compose up --wait`

**MediaWiki Major Version Upgrade (future)**

In a subsequent iteration we could make sure to always populate a build-specific version number in the `wikibase-config` volume, and check against this value on each boot of the container. When it diverges by a major version then the steps above would happen in a more automated way. For instance the current `wikibase-config` files could be backed up (this may be something we do automatically any on any conta)

## `wikibase-config` volume handling

A new optional `/config` volume mapping will be introduced to the Wikibase container. The `LocalSettings.*.php` files found there will be copied to/from `/var/www/html`. When the directory doesn't exists nothing will happen. When the directory does exists (because it was volume mapped in), and it is empty then the `/var/www/html/LocalSettings.*.php` files will be copied into it. If files already exists there then those files will be copied in / over what is in `/var/www/html`.

To reset or upgrade and installation a user would need to backup and then delete the files in `wikibase-config` (the presumed volume name), and then transcribe the settings back from their original configuration. However the values set in `SETUP_*` will again apply in the `LocalSettings.generated.php` and `LocalSettings.wbs.php` generation.

## `LocalSettings.php` architecture

Boot using `LocalSettings.generated.php`, `LocalSettings.wbs.php`, and `LocalSettings.override.php` (a static `LocalSettings.php` in Wikibase Image will include all 3 in this sequence)

1) `/var/www/html/LocalSettings.php`, runs the 3 LocalSettings files below in sequence
2) `/var/www/html/LocalSettings.generated.php` is the result of running the MediaWiki installer with the parameters we provide (all `SETUP_*` values)
3) `/var/www/html/LocalSettings.wbs.php` runs all WikibaseBundle extensions and other code specific to the WBS setup
4) `/wikibase-config/LocalSettings.php` is optional override via a volume mapping. If this file exists it will be ran after the above two files. Some default content may be provided with comments about some LocalSettings values to pay attention to  or set.

## Pseudocode for Wikibase `entrypoint.sh`

```
if /wikibase-config directory exists:
  if LocalSettings.php, etc exists in /wikibase-config:
    copy into /var/www/html
  else:
    if /var/www/html/LocalSettings.php, etc exists:
      copy it to /wikibase-config
else:
  do nothing

if /var/www/html/LocalSettings.php does not exist:
  run.php install/update process

start apache
```

## Some Open Questions

- `wgServer` in localhost testing? If set to an invalid URL (like the sample `template.env` values) then it will not work as it will try and redirect to this url on entry? Maybe. This may only be happening if `https` is specified as it seems MediaWiki by default will re-forward to HTTPS if accessed via HTTP when an HTTPS URL is provided for `wgServer`. If we stick with a protcol agnostic format here, which is supported, then I think that auto-redirect won't happen and less/nothing will break in a local installation when a `WIKIBASE_PUBLIC_HOST` value is set that is not yet accessible.
- General audit/review of http/https handling
- If LocalSettings doesn't exist and we re-run installer but database still exists, then notate which `SETUP_*` values will not be reset (or will they be?)
- Quickstatements host name change probably needs to re-run Quickstatements OAuth key generator in Wikibase image... Make this work appropriately without having to re-install, test, and document
- Explore all test suites without Elasticsearch and QuickStatements for speed, but keep Example test suite to have those
- Rename `example` to `wbs`
- Merge Wikibase and WikibaseBundle images
- On a upgrade / re-installation pay attention to `MW_SECRET_KEY`

.github/workflows/_build_test.yml
```diff
@@ -107,50 +107,12 @@ jobs:
           name: TestResults
           path: test/suites/**/results
 
-  test-upgrade:
```
rti, line 110: Would the recreation of upgrade tests be worth a ticket already?

RickiJay-WMDE: What's the reasoning behind removing them? My understanding was the work of moving from the `wikibase-shared` volume to a more limited `wikibase-config` was that that would *enable* proper upgrading, because the mw code would no longer be in a volume that persisted.

lorenjohnson: @rti it wouldn't bother me if we did. I had added a note to https://phabricator.wikimedia.org/T345224  as sorta a "heads-up" that a zoomed-out look at our testing paradigm (landscape) might be coming up soon. I think that is enough. 

@RickiJay-WMDE It's complicated, but the short story is the WIkibase image is pretty different now and the upgrade picture reflected in the Upgrade tests suite (as well as the docs), isn't a functional reflection of what upgrades look like going forward in this branch.

build/WDQS-frontend/README.md
```diff
@@ -8,6 +8,6 @@ Frontend for the Wikibase Query Service, as seen at [https://query.wikidata.org]
 | --- | --- | --- |
 | `LANGUAGE` | "en" | Language to use in the UI |
 | `BRAND_TITLE` | "DockerWikibaseQueryService" | Name to display on the UI |
-| `WIKIBASE_HOST` | "wikibase.svc" | Hostname of the Wikibase host |
-| `WDQS_HOST` | "wdqs-proxy.svc" | Hostname of the WDQS host (probably READONLY, hence use of the wdqs-proxy service) |
+| `WIKIBASE_HOST` | "wikibase" | Hostname of the Wikibase host (required) |
+| `WDQS_HOST` | "wdqs" | Hostname of the WDQS host (required). IT probably READONLY, hence use of the wdqs-proxy service) |
```
rti, line 12: >  IT probably READONLY, hence use of the wdqs-proxy service)

I do not understand what this wants to say.

build/Wikibase/LocalSettings.wikibase.php
```diff
@@ -0,0 +1,39 @@
...
+
+# Wikibase Repository
+wfLoadExtension( 'WikibaseRepository', "$IP/extensions/Wikibase/extension-repo.json" );
+require_once "$IP/extensions/Wikibase/repo/ExampleSettings.php";
+
+# Wikibase Client
+wfLoadExtension( 'WikibaseClient', "$IP/extensions/Wikibase/extension-client.json" );
+require_once "$IP/extensions/Wikibase/client/ExampleSettings.php";
+
+# Load extensions if present
```
rti, line 35:
```suggestion
# Load extensions if present, alphabetically ordered by filename
```

build/Wikibase/README.md
```diff
@@ -27,12 +29,11 @@ Wikibase running on MediaWiki.
 | `/var/www/html` | Base MediaWiki directory |
 | `/var/www/html/skins` | MediaWiki skins directory |
 | `/var/www/html/extensions` | MediaWiki extensions directory |
-| `/var/www/html/LocalSettings.d` | LocalSettings snippet directory. All PHP files from here will be loaded at the end of the default `/LocalSettings.php.template` |
+| `/var/www/html/LocalSettings.d` | LocalSettings snippet directory. All PHP files from here will be loaded at the end of other install generated settings, but before the settings in `/config/LocalSettings.override.php`
```
rti, line 32:
```suggestion
| `/var/www/html/LocalSettings.d` | LocalSettings snippet directory. All PHP files from here will be loaded in alphabetical order at the end of other install generated settings, but before the settings in `/config/LocalSettings.override.php`
```

build/Wikibase/entrypoint.sh
```diff
@@ -1,58 +1,69 @@
...
+        "$SETUP_MW_ADMIN_NAME"
+    # Include WBS customizations to generated LocalSettings.php
+    echo 'include "/var/www/html/LocalSettings.wikibase.php";' >> /var/www/html/LocalSettings.php
+    # Replace /config/LocalSettings.php with newly generated LocalSettings.php
+    cp /var/www/html/LocalSettings.php /config/LocalSettings.php
+    # Update the MW Admin email address (if this admin user doesn't already exist, a new one will be created)
+    php /var/www/html/maintenance/run.php resetUserEmail --no-reset-password "$SETUP_MW_ADMIN_NAME" "$SETUP_MW_ADMIN_EMAIL"
+fi
+
+# Always run update
```
rti, line 56:
```suggestion
# Always run update (this might be the first run off of a new image version on existing config and data)
```

build/WikibaseBundle/default-extra-install.sh
```diff
@@ -1,33 +1,35 @@
...
-set -ex
+set -x
 
 # Enables and configures elasticsearch index
-if [ -z "${MW_ELASTIC_HOST:-}" ] || [ -z "${MW_ELASTIC_PORT:-}" ] ; then
+if [ -z "${MW_ELASTIC_HOST:-}" ] ; then
     echo "Skipping Elasticsearch setup ..."
 else
-    /wait-for-it.sh "$MW_ELASTIC_HOST:$MW_ELASTIC_PORT" -t 300
+    /wait-for-it.sh "$MW_ELASTIC_HOST:${MW_ELASTIC_PORT}" -t 300
```
rti, line 8:
```suggestion
    /wait-for-it.sh "${MW_ELASTIC_HOST}:${MW_ELASTIC_PORT}" -t 300
```

test/suites/docker-compose.override.yml
```diff
@@ -6,5 +6,9 @@ networks:
 services:
   wikibase:
     volumes:
-      - ./LocalSettings.php:/var/www/html/LocalSettings.d/LocalSettings.test.php
+      - ./LocalSettings.php:/var/www/html/LocalSettings.d/ZZ-LocalSettings.test.php
```
rti, line 9: ❤️

build/WDQS/README.md
```diff
@@ -25,15 +25,15 @@ If you can not use [RecentChanges] then you will need to reload from an RDF dump
 
 | Variable | Default | Since | Description |
 | --- | --- | --- | --- |
-| `MEMORY` | "" | 0.2.5 | Memory limit for Blazegraph |
-| `HEAP_SIZE` | "1g" | 0.2.5 | Heap size for Blazegraph |
-| `WIKIBASE_HOST` | "wikibase.svc" | 0.2.5 | Hostname of the Wikibase host |
+| `WIKIBASE_HOST` | "wikibase" | 0.2.5 | Hostname of the Wikibase host |
 | `WIKIBASE_SCHEME` | "http" | 0.2.5 | Scheme of the Wikibase host |
-| `WDQS_HOST` | "wdqs.svc" | 0.2.5 | Hostname of the WDQS host (this service) |
+| `WDQS_HOST` | "wdqs" | 0.2.5 | Hostname of the WDQS host (this service) |
```
lorenjohnson, line 30: Previously the stated default for this, and the above WIkibase host, were not actually set to any default. No actual defaults are set and the defaults now reflect our Example compose setup which no longer sets up ".svc" aliases, but simply uses the name of the service (i.e. `wikibase`, `wdqs`, etc).

build/WDQS-frontend/README.md
lorenjohnson: Me neither. This is in the diff due to change of the default value, but the docs/txt was not update. This should be reviewed and updated along with the more comprehensive documentation update and review.

build/Wikibase/entrypoint.sh
```diff
@@ -1,59 +1,90 @@
...
 
-    # Copy our LocalSettings into place after install from the template
-    # https://stackoverflow.com/a/24964089/4746236
-    export DOLLAR='$'
-    envsubst < /LocalSettings.php.template > /var/www/html/LocalSettings.php
+    # Include WBS customizations to generated LocalSettings.php
+    {
+        set +u
+        if [ -n "$ELASTICSEARCH_HOST" ]; then
+            echo "\$elasticsearchHost = '$ELASTICSEARCH_HOST';"
```
lorenjohnson, line 60: This is instead of using an `envsubst` "template". This is to force `ELASTICSSEARCH_HOST` to be set statically in `LocalSettings.php` after initial install so it operates the same as the rest of the container environment variables. This creates predictable behaviour within the container and makes the `config/LocalSettings.php` generated after initial container run the single source of configuration for the container.

build/Wikibase/entrypoint.sh
```diff
@@ -1,59 +1,90 @@
...
 
-    # Run extrascripts on first run
     if [ -f /extra-install.sh ]; then
         # shellcheck disable=SC1091
-        source /extra-install.sh
+        bash /extra-install.sh
     fi
 fi
 
-# Copy LocalSettings.php to a shared directory, if the image is being used with this shared directory existing
```
lorenjohnson, line 49: This whole way of doing things is now replaced by having a required `config` volume.

build/Wikibase/jobrunner-entrypoint.sh
```diff
@@ -3,25 +3,13 @@
...
-sleep 1
-/wait-for-it.sh "$DB_SERVER" -t 300
-
 kill_runner() {
 	kill "$PID" 2> /dev/null
 }
 trap kill_runner SIGTERM
 
 while true; do
-	if [ -e /shared/LocalSettings.php ]; then
```
lorenjohnson, line 19: `/config/LocalSettings.php` is now required in the entry point for the container to run, so this check is no longer needed.

example/docker-compose.yml
```diff
@@ -1,17 +1,5 @@
 name: wikibase-suite-example
 
-x-common-variables: &wikibase_variables
```
lorenjohnson, line 3: This extra common Yaml variable no longer need to be created as these parameters are passed only to the `wikibase` service which then generates a `LocalSettings.php` on first run. `wikibase-jobrunner` also uses that volume so gets all its settings then from the `LocalSettings.php` file.

.github/workflows/_build_test.yml
adeeri: If they no longer make sense in their current form, should the upgrade functionality be tested some other way?

lorenjohnson: Yes, for sure. Here is the draft ticket created for that during Sprint Planning today: https://phabricator.wikimedia.org/T364847.

build/WDQS/README.md
rti: could be resolved IMHO

build/Wikibase/entrypoint.sh
rti: could be marked as resolved for now imho

test/specs/repo/extensions/wikibase-local-media.ts
```diff
@@ -76,10 +77,6 @@ describe( 'WikibaseLocalMedia', function () {
 		await propertyIdSelector( propertyId ).click();
 
 		await browser.keys( 'image.png'.split( '' ) );
-		await $( 'ul.ui-mediasuggester-list' )
-			.$( 'a' )
-			.$( 'span[title="Image.png"]' )
-			.click();
```
rti, line 82: should this really be changed as part of this PR?

build/Wikibase/LocalSettings.wbs-extensions.php
```diff
@@ -0,0 +1,36 @@
+<?php
```
rti, line 1: should we maybe just call it `LocalSettings.wbs.php`? It is doing more than just the extensions.

example/docker-compose.yml
```diff
@@ -24,41 +12,38 @@ services:
         condition: service_healthy
     restart: unless-stopped
     ports:
-      - ${WIKIBASE_PORT}:80
+      - ${WIKIBASE_PORT:-8880}:80
```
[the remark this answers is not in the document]
rti: closing this here to move the PR forward. follow up ticket is https://phabricator.wikimedia.org/T365038

rti (changes requested):
see my previous comments

test/specs/repo/extensions/nuke.ts
```diff
@@ -38,16 +38,12 @@ describe( 'Nuke', function () {
 
 		await browser.acceptAlert();
 
-		await $( 'li*=has been queued for deletion' );
-
-		await browser.waitForJobs();
```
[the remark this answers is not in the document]
lorenjohnson: The test was test was split into two, but the same thing is being tested. The "confirm that was actually deleted" part had to be moved to a separate test so it could have `.retries` applied as it was being flakey now that things are running in Job Runner.

test/specs/repo/extensions/wikibase-local-media.ts
lorenjohnson: It had to be to make the tests pass. Due to additional latencies with the Job Runner based config it wouldn't pass reliably as it was written.

example/docker-compose.yml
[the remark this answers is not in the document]
lorenjohnson: I am all for removing these, and will do it in this branch now. I totally agree and think this will be a lot more clear. A lot of these changes were happening in somewhat conservative steps.  Then things like this become revealed, and I'm like-- oh yeah--that can go away entirely now 👍🏼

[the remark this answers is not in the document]
lorenjohnson: Done ✔️

build/Wikibase/LocalSettings.wbs-extensions.php
lorenjohnson: agreed, and updated.

build/WikibaseBundle/default-extra-install.sh
```diff
@@ -1,33 +1,35 @@
 #!/usr/bin/env bash
-set -ex
+set -x
```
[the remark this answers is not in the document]
lorenjohnson: This one was tricky and took several hours to get to work right. The issue was the hard exit that the QuickStatements OAuth key generator would throw on re-run would cause a cascade of failure.  It might however be ok now that this is only rant on "install" now and not on every boot. Will try again...

test/specs/repo/extensions/nuke.ts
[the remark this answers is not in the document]
rti: Got it. My impression is, that relying on retries instead of explicitly waiting for jobs to be finished is quite implicit and hard to understand intuitively. Not a blocker for the PR though.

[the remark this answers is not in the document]
lorenjohnson: Of course I agree, I can say more :)

Despite that I think this becoming flakey somehow had to do with the Job Runner running instead of jobs on request, `waitForJobs` then checking was still not giving a consistent result. I had watched it headed and spent a couple hours going over different options before I landed on this. It simply is not that predictable when the page will disappear and we don't seem to have a way to track when it's done, so it came down to retires or a long `browser.pause()`, and I chose the later.  `retries` I think is the better of those two options, and I think everywhere else it is applied currently in our code is the same situation: there wasn't another clear way and it didn't come down to jobs clearing the queue...

test/specs/repo/extensions/wikibase-local-media.ts
lorenjohnson: Also -- to be clear, this change and the test change above, were both flakey at the time of merging from the "tests run against Example" PR, they were just further fixed here.

rti: Got it

rti (approved):
## Awesome 🚀🚀🚀

PR #668, opened 2024-04-25
Update example

adeeri:


rti (approved):

lorenjohnson (approved):
LGTM. To this point we have made this change after release so we can see tests pass, but I do understand that the Dockerhub release is being held back until a release announcement is sent.

PR #675, opened 2024-05-15
T365141, T364956: Change default to HTTPS in example and uploads on for Wikibase. Set max file upload size in parity with MediaWiki, change to Traefik from nginx-proxy.

lorenjohnson:
* Makes SSL / HTTPS and ACME certificate generation the default for Example: https://phabricator.wikimedia.org/T365141

* Addresses the file upload size issue detailed here: https://phabricator.wikimedia.org/T364956

* Sets `$wgEnableUploads` on by default which is divergent from the MediaWiki stock installation which has it off by default. 

* Moves to Traefik as our reverse proxy and SSL provisioner as it removes the nginx-proxy file upload limits as yet another place to have to set this limit. Also, while a bit more complex, in testing Traefik configuration  feels a _little bit_ more clear than the `nginx-proxy`  and `acme-companion` setup overall. 

NOT done in this branch:

* A `example/config` directory with sample config files like `config/LocalSettings.override.php`, `config/wikibase-php.ini`, etc. I have not done this yet here as it isn't strictly required if we just go with allowing a 100M default max upload setting, and have uploads on by default. I think making a static set of of `example/config` files can be another PR and further considered...


example/docker-compose.yml
```diff
@@ -177,29 +190,29 @@ services:
   # C. REVERSE PROXY AND SSL SERVICES
   # --------------------------------------------------
 
-  # See https://github.com/nginx-proxy/nginx-proxy/blob/main/docs/README.md for setup and options
-  nginx-proxy:
-    image: nginxproxy/nginx-proxy
+  traefik:
+    image: traefik:v2.5
+    command:
+      - "--api.insecure=true"
```
rti, line 196:
```suggestion
      #- "--api.insecure=true"
```
I think this should be off by default.

rti (approved):
### Approved besides my two suggestions above.

> Sets $wgEnableUploads on by default which is divergent from the MediaWiki stock installation which has it off by default. Should we?

I like that. Let's go with that for now.

> Still leave SSL off by default as services are inaccessible on HTTP locally when certs can't be generated otherwise. This was more or less true with nginx-proxy / acme-companion as well, which is why I had Acme commented out by default there before as well.

I just tried to turn https on on a vps with properly set up dns a records and found me fiddling with too many details. In a follow up task I would like to explore how https by default would look like, especially, how bad the failure in a "non publicly routable address" case would look like. My guts is: it does not fail too bad, usable enough for a "local" test (accept invalid certs in the browser). Otherwise we want to primarily support public production hosts anyway. https://phabricator.wikimedia.org/T365141

> Re. making a example/config directory with sample config files like config/LocalSettings.override.php, config/wikibase-php.ini, etc. I have not done this yet here as it isn't strictly required if we just go with allowing a 100M default max upload setting, and have uploads on by default. I think making a static set of of example/config files can be another PR and further considered...

Created a follow up task https://phabricator.wikimedia.org/T365135

lorenjohnson:
> ### Approved besides my two suggestions above.
> > Sets $wgEnableUploads on by default which is divergent from the MediaWiki stock installation which has it off by default. Should we?
> 
> I like that. Let's go with that for now.
> 
> > Still leave SSL off by default as services are inaccessible on HTTP locally when certs can't be generated otherwise. This was more or less true with nginx-proxy / acme-companion as well, which is why I had Acme commented out by default there before as well.
> 
> I just tried to turn https on on a vps with properly set up dns a records and found me fiddling with too many details. In a follow up task I would like to explore how https by default would look like, especially, how bad the failure in a "non publicly routable address" case would look like. My guts is: it does not fail too bad, usable enough for a "local" test (accept invalid certs in the browser). Otherwise we want to primarily support public production hosts anyway. https://phabricator.wikimedia.org/T365141
> 
> > Re. making a example/config directory with sample config files like config/LocalSettings.override.php, config/wikibase-php.ini, etc. I have not done this yet here as it isn't strictly required if we just go with allowing a 100M default max upload setting, and have uploads on by default. I think making a static set of of example/config files can be another PR and further considered...
> 
> Created a follow up task https://phabricator.wikimedia.org/T365135

I agreed with all of that, so let's just make it on by default and then work back from there to a better "local" testing scenario? I am going to go ahead and make that change and flag for re-review to make sure you're good with it. 

The thing is whatever we do here will want some degree of documentation, so I would rather go with what makes sense from there. 

I've been testing on my own VPS as well and the ACME generated certificates are working well on first try there.

lorenjohnson:
@rti I'm also curious if you are a fan of this move to Traefik from nginx-proxy? I think I prefer it. It feels like a more "mature" solution. Do you think it's better we do that or not?

rti (approved):
Boom. https out of the box. Nice :)

PR #682, opened 2024-05-23
T366041: Example config in directory

rti:
https://phabricator.wikimedia.org/T366041

Note: the branch name carries the wrong ticket number


build/Wikibase/LocalSettings.wbs.php
```diff
@@ -18,21 +18,22 @@
 
 $wgArticlePath = "/wiki/$1";
 
-# Wikibase Repository
```
lorenjohnson, line 21: Love to see this moved into the LocalSettings.d 🙌🏼👏🏼🎉

test/suites/client-LocalSettings.php
```diff
@@ -1,39 +1,13 @@
...
-$wgScriptPath = "/w";        // this should already have been configured this way
-$wgArticlePath = "/wiki/$1";
-
-wfLoadSkin( 'Vector' );
-
 $wgEnableWikibaseRepo = false;
 $wgEnableWikibaseClient = true;
 
+# use jobs on request, client does not have a job runner
+$wgJobRunRate = 1;
```
lorenjohnson, line 9: Good catch! How was this working before?

test/suites/client-LocalSettings.php
```diff
@@ -57,32 +31,3 @@
 $wgWBClientSettings['siteLinkGroups'] = [ 'mywikigroup' ];
 $wgWBClientSettings['siteGlobalID'] = 'client_wiki';
 
-foreach (glob("LocalSettings.d/*.php") as $filename)
```
lorenjohnson, line 60: Yay for making this all go away! So this list was comprehensive excluding our added extensions before, and it's ok that no extensions are being added now?

test/suites/repo-LocalSettings.php
```diff
@@ -1,45 +1,12 @@
...
 $wgEnableWikibaseRepo = true;
 $wgEnableWikibaseClient = false;
 
 wfLoadExtension( 'WikibaseRepository', "$IP/extensions/Wikibase/extension-repo.json" );
 require_once "$IP/extensions/Wikibase/repo/ExampleSettings.php";
+
 $wgWBRepoSettings['siteLinkGroups'] = [ 'mywikigroup' ];
 $wgLocalDatabases = $wgWBRepoSettings['localClientDatabases'] = [ 'client_wiki', ];
-
-foreach (glob("LocalSettings.d/*.php") as $filename)
```
lorenjohnson, line 42: Same question as above. Happy to see this go away, but also wondering if we're testing anything different now that some extensions that were loaded before are not longer being loaded?  Not terribly concerned either way, but curious.

lorenjohnson (approved):
Looks good to me! Really good calls and clean-ups! Happy for this all going forward 🥳

test/suites/client-LocalSettings.php
rti: Well, `1` is the default value for `wgJobRunRate`. This file here was overriding `LocalSettings.wbs.php` for the client instance before. So on client `wgJobRunRate` was never set.

Now, this file here does not become the `LocalSettings.wbs.php` in the client, but gets mounted to `LocalSettings.d/20_Wikibase.php`, overriding the original wikibase loading only. So the original `LocalSettings.wbs.php` is still in client, sets `wgJobRunRate` to `0`, then loads all the extensions from `LocalSettings.d` and therefore sets `wgJobRunRate` to `1` in the client case.

PR #683, opened 2024-05-23
T362226: example docs (1st review)

rti:
depends on #682


deploy/config/README.md
```diff
@@ -0,0 +1,5 @@
+# Wikibase Suite Deployment Kit Configuration
+
+Will directory will contain your configuration files after the first run of the Wikibase Suite Deployment Kit. Feel free to make adjustments as you see fit. Backup this folder regularly in order to restore your configuration later.
```
lorenjohnson, line 3: "Will directory will contain" to "Directory wil contain"

lorenjohnson, line 3: "Will directory will contain" to "Directory wil contain"

deploy/config/README.md
```diff
@@ -0,0 +1,5 @@
...
+
+Find information on how to run the Wikibase Suite Deployment Kit in the top level ../README.md
```
lorenjohnson, line 5: I like how tidy it is to have the docs all in the root README, but I don't like forcing consumers of WBS to have to wade through the build and test related docs on first contact with the product. I think I still see deploy as an independent product that the Build Pipeline consumes, whether it is is in its own repo yet or not. TBD.

deploy/docker-compose.yml
```diff
@@ -190,18 +198,18 @@ services:
       - "--entrypoints.web.http.redirections.entryPoint.scheme=https"
       - "--entrypoints.web.http.redirections.entrypoint.permanent=true"
       # ACME SSL certificate generation
-      - "--certificatesresolvers.myresolver.acme.httpchallenge=true"
-      - "--certificatesresolvers.myresolver.acme.httpchallenge.entrypoint=web"
-      - "--certificatesresolvers.myresolver.acme.email=${MW_ADMIN_EMAIL}"
-      - "--certificatesresolvers.myresolver.acme.storage=/letsencrypt/acme.json"
+      - "--certificatesresolvers.letsencrypt.acme.httpchallenge=true"
```
lorenjohnson, line 201: How wow, how was this working at all while it was still set to "myresolver"?

deploy/template.env
```diff
@@ -0,0 +1,37 @@
...
+# Compose setup.
+#
+# WARNING: Do not add comments on the same line as env vars, as in some
+# environments they will be included in the var!
+#
+# shellcheck disable=2034 # do not warn about unused variables
+# ##############################################################################
+
+# Public Hostname Configuration.
+# Define the DNS Domain Names for your Wikibase Suite services. The DNS entries
```
lorenjohnson, line 14: Suggested change to passive voice, and somewhat more clear:

"These domain names for your Wikibase Suites services should be configured on your DNS server to point to the public IP address of the server being deployed to."

deploy/template.env
```diff
@@ -0,0 +1,37 @@
+# ##############################################################################
+# Wikibase Suite
+#
+# This file is a template for the .env file used by the Wikibase Suite Docker
+# Compose setup.
+#
+# WARNING: Do not add comments on the same line as env vars, as in some
+# environments they will be included in the var!
+#
+# shellcheck disable=2034 # do not warn about unused variables
```
lorenjohnson, line 10: I know you had commented on this before, but just checking: is there no place we can configure this exception exception inline?

README.md
```diff
@@ -44,10 +47,11 @@ $ ./test.sh repo --spec specs/repo/special-item.ts
 $ ./test.sh repo --setup
 ```
 
-### Example Instance Commands
+### Deploy
 
 ```
-$ docker compose --env-file ./example/template.env up --wait
+$ cd deploy
```
lorenjohnson, line 53: I may have missed it elsewhere, but I'm not seeing where we guide the user to cp `template.env` to `.env`. Also, as mentioned in my earlier comment, I would still prefer to keep these docs in the `deploy` directory. 

Beyond those two things though, I appreciate the brevity here and which I think encourages users to look more careful at comments in the configuration files.

lorenjohnson (changes requested):
Lots of great things! It feels sorted and close. Other than the "deploy" naming, which I will address in the related Phab ticket, My biggest concerns are about the location of the Deploy docs in the root readme, and about the "Deploy" naming.

deploy/config/README.md
lorenjohnson: Disregard the part about the location of the doc. I hadn't seen the whole document and thought the diff here was it. Sorry!

deploy/README.md
```diff
@@ -0,0 +1,309 @@
...
+
+This WBS Deployment Kit uses the following WBS Service Images, all of them packaged as Docker Images:
+
+- **[Wikibase](https://hub.docker.com/r/wikibase/wikibase)** MediaWiki packaged with the Wikibase extension and other commonly used extensions.
+- **Job Runner** The MediaWiki [JobRunner](https://www.mediawiki.org/wiki/Manual:Job_queue#Cron) service which uses the same Wikibase container as above.
+- **[MariaDB](https://hub.docker.com/_/mariadb)** Database service for MediaWiki and Wikibase.
+- **[Elasticsearch](https://hub.docker.com/r/wikibase/elasticsearch)** Search service used by MediaWiki.
+- **[WDQS](https://hub.docker.com/r/wikibase/wdqs)** Wikidata Query Service allowing SPARQL queries.
+- **[WDQS Frontend](https://hub.docker.com/r/wikibase/wdqs-frontend)** Web Frontend to run SPARQL queries.
+- **[wdqs-proxy](https://hub.docker.com/r/wikibase/wdqs-proxy)** A middle layer for WDQS which serves to filter requests and make the service more secure.
```
lorenjohnson, line 19: Title case "WDQS Proxy" to be more inline with the other naming here?

deploy/README.md
```diff
@@ -0,0 +1,309 @@
...
+#### Hardware
+
+- Network connection with a public IP address
+- AMD64 architecture
+- 8 GB RAM
+- 4 GB free disk space
+- Docker 22.0, or greater
+- Docker Compose 2.10, or greater
+
+#### Domain Names
```
lorenjohnson, line 39: I think this section could/should be a section within "Initial Setup"

deploy/README.md
```diff
@@ -0,0 +1,309 @@
...
+🎉 Congratulations, your Wikibase Suite instance should be up and running. Web
+interfaces are available via HTTPS (port 443) on the domain names you
+configured for Wikibase, WDQS Frontend and Quickstatements.
+
+#### Stopping
+
+To stop, use 
+
+```sh
+docker compose down
```
lorenjohnson, line 90: I think this should be `docker compose stop` and `down` reserved for the updating process detailed below.

deploy/README.md
```diff
@@ -0,0 +1,309 @@
...
+docker compose up --wait
+```
+
+### Advanced Configuration
+On first launch, WBS Deployment Kit will create files in the `./config` directory next to your `.env` file and the `docker-compose.yml` and `template.env`. This is the configuration of your instance. **You own those files.** Be sure to include them in your backups.
+
+#### `config/LocalSettings.php`
+This file is generated by the [MediaWiki installer script](https://www.mediawiki.org/wiki/Manual:Install.php) and augmented by the Wikibase containers `entrypoint.sh` on first launch. Once this file has been generated, you own it. This means you can make changes, you even might need to make changes on major version updates (as describes in MediaWiki UPGRADE notes). In order to ease keeping track of your own adjustments vs settings generated by the Wikibase container, we provide a file called `LocalSettings.override.php`. This file is the preferred location for your adjustments. See below.
```
lorenjohnson, line 115: Worth considering making "UPGRADE" and similar be in-document bookmarks. Not sure how well that works on Github, but I think I've seen it used: https://stackoverflow.com/a/7335259

deploy/README.md
```diff
@@ -0,0 +1,309 @@
...
+
+The absence of `LocalSettings.php` is the trigger for the Wikibase container to run the [MediaWiki installer script](https://www.mediawiki.org/wiki/Manual:Install.php). If you need to run the installer again, remove the generated `LocalSettings.php` file and restart your instance.
+
+#### `config/LocalSettings.override.php`
+This file is the location for your MediaWiki and Wikibase customizations. If you need to regenerate your `LocalSettings.php`, your `LocalSettings.override.php` is the safe place that is never rewritten, even when running the [MediaWiki installer script](https://www.mediawiki.org/wiki/Manual:Install.php).
```
lorenjohnson, line 120: Optional: Note that settings in the override file happen after extensions are loaded?

deploy/README.md
```diff
@@ -0,0 +1,309 @@
...
+Restart your Wikibase container with 
+```sh
+docker restart wikibase-suite-wikibase-1
+```
+
+#### `config/wikibase-php.ini`
+This file is a `php.ini` override file, a good place to tune PHP configuration values. It will be loaded by the Wikibase Webservers PHP interpreter.
+
+#### docker-compose.yml
+To further customize your instance, consider adjusting `docker-compose.yml`. In order to ease updating to newer versions of the WBS Deployment Kit, consider putting your customization into a new file called `docker-compose.override.yml` and from now on, restart your instance with
```
lorenjohnson, line 137: with:

deploy/README.md
```diff
@@ -0,0 +1,309 @@
...
+
+```sh
+docker compose -f docker-compose.yml -f docker-compose.override.yml down
+docker compose -f docker-compose.yml -f docker-compose.override.yml up --wait
+```
+
+This way your changes are kept separate from the original WBS Deploying Kit code.
+
+### Managing your Data
```
lorenjohnson, line 146: Nitpick and a matter of opinion for sure, but I wouldn't capitalise "Data" here or below.

deploy/README.md
```diff
@@ -0,0 +1,309 @@
...
+
+### Updating
+
+WBS is versioned with [Semantic Versioning](https://semver.org/spec/v2.0.0.html). The WBS Deployment Kit and all the WBS Service Images have individual version numbers.
+
+> 💡 WBS Deployment Kit always references the latest minor and patch releases of the compatible WBS Service Images major versions.
+
+#### Minor and patch updates for WBS Service Images
+
+As WBS Deployment Kit always references latest minor and patch releases of compatible WBS Service Images TODO: rephrase, non breaking changes (including security updates) are applied automatically when recreating Docker containers. This should be always safe to do. Simply run
```
lorenjohnson, line 201: TODO...

deploy/README.md
```diff
@@ -0,0 +1,309 @@
...
+
+```sh
+docker compose down 
+docker compose up --wait
+```
+
+> 💡 If you do not want to pull in new versions of WBS Service Images on container restart, just stop your containers using `docker compose stop` before restart. This will keep the current containers intact. Please note that this will not apply any security updates. It is generally recommended to remove your containers for restart with `docker compose down`.
```
lorenjohnson, line 208: Recommending using `down` as general practice is likely a good call, but I'm not certain about it. Do you feel sure about that?

deploy/README.md
```diff
@@ -0,0 +1,309 @@
...
+
+#### Minor and patch updates for WBS Deployment Kit
+
+WBS Deployment Kit major versions are tracked in dedicated branches such as `TODO:BRANCH_NAME_HERE`. Therefor, pulling from such a branch will only update minor and patch version but never trigger breaking changes.
```
lorenjohnson, line 212: "Therefor" > "Therefore"

deploy/README.md
```diff
@@ -0,0 +1,309 @@
...
+
+If you did not change `docker-compose.yml`, you can update simply using `git pull`.
```
lorenjohnson, line 214: In the case of minor and patch updates I don't think we make changes to `deploy/docker-compose.yml`

deploy/README.md
```diff
@@ -0,0 +1,309 @@
...
+# TODO check BRANCH/TAG before release
+git pull
+```
+If you have made complex changes to `docker-compose.yml`, `git commit` them to a separate branch and `git merge` them with upstream changes as you see fit.
+
+> 💡 One major version of WBS Deployment Kit will always reference the only one major version for each of the WBS Service Images. Therefor, updating WBS Deployment Kit minor and patch versions from its major versions git branch will never lead to breaking changes in WBS Service Images. Those updates are always considered safe!
+
+#### Major upgrades
+
+Major version upgrades are performed by updating WBS Deployment Kits major version. This might reference new major versions of WBS Service Images too and therefor include breaking changes. Those may require additional steps. Refer to specific upgrade instructions in the [Migration Guide](#migration-guide) section below.
```
lorenjohnson, line 226: "therefor" > "therefore"

deploy/README.md
```diff
@@ -0,0 +1,309 @@
...
+
+WBS only supports updating from one major version to the next one. If you want to upgrade from 21 to 23, you need to upgrade from 21 to 22 first and in a next step from 22 to 23.
+
+> ‼️ Always [create a backup](#backup-your-data) of your data before performing a major version update.
```
lorenjohnson, line 230: "Also backup changes to `docker-compose.yml` and look for changes in the `template.env` file."

deploy/README.md
```diff
@@ -0,0 +1,309 @@
...
+
+Once all version specific migration activities are done, use `git` to update your WBS Deployment Kit to the new version.
+
+```sh
+docker compose down 
+git remote update
+git checkout NEWVERSIONBRANCH
```
lorenjohnson, line 237: `NEWVERSIONTAG` ?

deploy/README.md
```diff
@@ -0,0 +1,309 @@
...
+git pull
+docker compose up --wait
+```
+
+#### Automatic updates
+At the moment, WBS Deployment Kit does not support automatic updates. To deploy minor and patch updates, including security fixes, [restart you instance](#minor-and-patch-updates-for-wbs-service-containers) on a regular basis, e.g. with a systemd timer or cron job.
```
lorenjohnson, line 243: Maybe this comment could be combined with the update instructions below or as part of the FAQ?

deploy/README.md
```diff
@@ -0,0 +1,309 @@
...
+‼️ **This will destroy all data, make sure to [backup anything](#backup-your-data) you wish to retain.**
+
+To reset the configuration and data, remove the Docker Containers, Docker Volumes and the generated `LocalSettings.php` file.
+
+```sh
+docker compose down --volumes
+rm config/LocalSettings.php
+```
+
+Removing the `traefik-letsencrypt-data` volume will request a new certificate from LetsEncrypt on next launch of your instance. Certificate generation on LetsEncrypt is [rate limited](https://letsencrypt.org/docs/rate-limits/). Eventually, you might be blocked from generating new certificates **for a couple of days**. To prevent that, switch to the LetsEncrypt staging server by appending the following to your `traefik` `command` in  your `docker-compose.yml` file:
```
lorenjohnson, line 259: As I think there is still a comment to this effect in the `docker-compose.yml`. I would slightly prefer it only being documented in one place. Perhaps we make this warning here and point to the `docker-compose.yml` for further instructions? Alternatively, maybe we can leave this as-is and remove the commented-out section in `docker-compose.yml` with a brief warning there pointing back to the README docs in its place. ?

deploy/README.md
```diff
@@ -0,0 +1,309 @@
...
+
+```sh
+# TODO check BRANCH/TAG before release
+git pull
+```
+If you have made complex changes to `docker-compose.yml`, `git commit` them to a separate branch and `git merge` them with upstream changes as you see fit.
+
+> 💡 One major version of WBS Deployment Kit will always reference the only one major version for each of the WBS Service Images. Therefor, updating WBS Deployment Kit minor and patch versions from its major versions git branch will never lead to breaking changes in WBS Service Images. Those updates are always considered safe!
+
+#### Major upgrades
```
lorenjohnson, line 224: I wonder if this may be better as the beginning of the the "Migration Guide" section?

lorenjohnson, line 224: I think this section should be combined with the Migration guide, and here should just be the note pointing there.

deploy/README.md
```diff
@@ -0,0 +1,309 @@
...
+
+### Can I host Wikibase Suite locally?
+
+Yes, Wikibase Suite can be hosted locally for testing purposes by using e.g. the example domain names `*.mydomain.net` from `template.env` in your `.env` file. Configure those domain names also in your `/etc/hosts` to make them resolve to `127.0.0.1`.
+
+However, due to OAuth requirements, QuickStatements may not function properly without a publicly accessible domain names for both the `WIKIBASE_PUBLIC_HOST` and `QUICKSTATEMENTS_PUBLIC_HOST`. Running locally without publicly accessible addresses will also not generate a valid SSL certificate so accessing services will require allowing the invalid certificate on first load in the browser. 
+
+### Do you have any recommended Internet host / VPS provider recommendations?
+
+At this time, there are no specific recommendations for Internet hosts or VPS providers for hosting Wikibase Suite. The suite has been tested on various providers, and as long as the minimum technical requirements are met, it should run as expected.
```
lorenjohnson, line 307: Maybe bookmark back to Requirements > Hardware above

lorenjohnson (dismissed):
Ok, sorry again for the extra noise. I've now deleted some initial comments that came from my confusion about the location of the `deploy/README.md`. I have now finished my notes on the `deploy/README.md`. Such good work! I feel like we'll have some rounds of refinement both before and after this next release, but it seems to me you covered a lot of the bases here! 🙌🏼

deploy/docker-compose.yml
```diff
@@ -190,18 +198,18 @@ services:
...
+      - "--certificatesresolvers.letsencrypt.acme.httpchallenge.entrypoint=web"
+      - "--certificatesresolvers.letsencrypt.acme.email=${MW_ADMIN_EMAIL}"
+      - "--certificatesresolvers.letsencrypt.acme.storage=/letsencrypt/acme.json"
       # Uncomment this line to only test ssl generation first, makes sures you don't run into letsencrypt limits
```
lorenjohnson, line 205: "makes sures" to "makes sure"

README.md
```diff
@@ -1,18 +1,21 @@
-# Wikibase release pipeline
+# Wikibase Suite
```
lorenjohnson, line 1: Wikibase Suite Release Pipeline

README.md
```diff
@@ -58,6 +62,113 @@ To take advantage of the git hooks we've included, you'll need to configure git
...
+
+[ ] **Agree with ComCom on a timing to publish the release**. Talk to ComCom and TechWriter about that.
+
+[ ] **Publish the release** by merging the release preparation branch into the release branch. ‼️ ‼️ ‼️ **This will automatically push docker images to docker hub and therefore release the new version!**
+
+[ ] **Update the [docker install instructions](https://www.mediawiki.org/wiki/Wikibase/Docker)** to reflect the latest version. If you made a bugfix release for an older release, this can be skipped.
+
+[ ] **Make sure the communication is sent.**
+
+[ ] **Update the deploy directory** ☠️☠️☠️ For this to happen, we currently need to merge another PR to the release branch. This PRs only change should be making the deploy directory reference the new version. For this PRs pipeline to properly run, you need to manually delete the git tag of your just released version from the github repo before merging. Merging this deploy directory update PR is technically a re-release. This weirdness will hopefully be fixed soon.
```
lorenjohnson, line 169: This I think is no longer true. It is possible to update to the expected release tags in deploy ahead of time as tests don't run against test builds only.

deploy/config/README.md
rti: Thanks, fixed

deploy/docker-compose.yml
rti: Well, it is just a name after all, but I thought this one might be more descriptive.

deploy/template.env
rti: sold

deploy/template.env
rti: Yeah, that probably really a config that belongs on my machine.

I should remove it from the file, right?

deploy/README.md
rti: done

deploy/README.md
```diff
@@ -0,0 +1,309 @@
...
+
+Most values set in `.env` are copied statically into the respective containers
+after the first time you run `docker compose up`.
+
+To reset the configuration but keep your existing data:
+
+1. Make any needed changes to the values in the `.env` file copied from
+   `template.env` above. NOTE: Do not change `DB_*` values unless you are also
+   re-creating the database ([see below](#removing-wikibase-suite-completely-with-all-its-data).
+2. Delete your LocalSettings.php from the ./config directory.
```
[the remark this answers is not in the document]
rti: Good catch. Done

deploy/README.md
rti: Done

deploy/README.md
rti: like it.
done

deploy/README.md
rti: went back and forth on that. trusting your guts. done.

deploy/README.md
darthmon: good question - how much docker knowledge are we assuming from the users here? If we assume they know the difference I would say: let's say sth like "use docker compose down if you want to remove your containers". Basically without recommending anything unless we think it is crucial for the good working of WBS. Does this resonate?

deploy/README.md
darthmon: placeholder?

deploy/README.md
rti: Well, considering the containers as "a stateless instance of the code" (that is the docker image), it should be safe to always throw them away. Practicing that could also be a way to ensure that it is safe. 

We have config in `./config` now and all data in docker volumes. So I would vote for removing containers and use `down`. Otherwise I would frighten we could hide a dependency on state in the container we do not expect.

deploy/README.md
rti: leaving it here as a reminder to myself

deploy/README.md
rti: I could think of cases. Maybe updating traefik from 2.5 to 2.6 could be one of them.

rti: I mean, technically we could even add features such as an reverse proxy, except that we did that already ^^

deploy/README.md
rti: done

deploy/README.md
rti: Yes, that's a placeholder.

Maybe we need to spend a moment with @adeeri on the drawing board again in order to get the full tag vs branch picture here.

deploy/README.md
rti: Good point, leaving it open as a reminder to review the major update procedure.

deploy/README.md
rti: I think it fits here quite well. Changed the text a bit. Could you have a look again please - and - if you still think it should be moved elaborate a bit why?

deploy/README.md
rti, line 1: TODO: add latest suggestions from https://docs.google.com/document/d/1TirRCqOvZUxS4p9YayOP-t1DG3j03b-Ml0ey6r5M_u0/edit?disco=AAABOCfaTpc&usp_dm=true before merge

test/specs/repo/queryservice.ts
```diff
@@ -277,9 +277,9 @@ describe( 'QueryService', function () {
 	} );
 
 	it( 'Should show results from a page in allowlist.txt', async function () {
-		// We don't currently have a way for the example suite to pass tests with breaking changes
+		// We don't currently have a way for the Deployment Kit to pass tests with breaking changes
```
darthmon, line 280: just an honest question to make sure: is shortening to "Deployment Kit" alright or should we stick to "WBS Deployment Kit" for coherence? Can there be any confusion at all?

darthmon (commented):
hey Robert, after checking the doc I reviewed the last commits. Great work!

deploy/README.md
lorenjohnson: Fair, I guess I was assuming we'd move up a major version for deploy kit at such changes, but technically yes we could add non-breaking configuration or additional services and consider them non-breaking/minor
updates...

deploy/README.md
lorenjohnson: In spirit I'm with Mónica on this one, but in practice I am ok with suggesting `down` as a general rule. I just think from a Docker perspective it is a little surprising, though not enough to warrant an explanation.

I feel there may be cases where re-provisioning the containers from scratch is ill advised, but other than it implying a longer restart cycle on the service, I can't think of any cases now.

However, to your point I do think we should be confident that our containers are stateless regardless, and that is something we should look to our test suite to adequately prove. Our images when used in another setup or orchestration layer (i.e. Kubernetes) may be stopped and started vs down'd as standard practice.

That's my thinking on this, but having said all that, I am ok with this as it is for now. Just things worth us keeping in mind.

deploy/README.md
lorenjohnson: In general I'm also always am back and forth re Title Case in docs. But in this instance I'm flagging it as Dan recently pointed out to me that the MediaWiki way is _never_ title case.   I may have gotten that wrong, and it certainly does that need to apply to our code comments, but may come-up when we look into how to sync the README file with the MediaWiki Wikibase Docker install page...

deploy/template.env
lorenjohnson: I think so, yes, but Spellcheck is part of our linting chain so think it will complain without it here or us disabling the rule elsewhere?

README.md
```diff
@@ -1,18 +1,21 @@
-# Wikibase release pipeline
+# Wikibase Suite
 
-## Repository overview
+Wikibase Suite (WBS) is a containerised, production ready [Wikibase](https://wikiba.se) system, that allows you to self host a knowledge graph similar to Wikidata.
```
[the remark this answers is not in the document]
rti: good idea. done

README.md
```diff
@@ -58,6 +62,113 @@ To take advantage of the git hooks we've included, you'll need to configure git
...
 ```
 
-## Documentation
+## Testing
 
-The pipeline documentation can be found [here](docs/index.md).
+Tests are organized in "suites" which can be found in `test/suites`. Each suite runs a series of specs (tests) found in the `test/specs` directory. Which specs run in each suite by default are specified in the `.config.ts` file in each suite directory under the `specs` key.
+
+All test suites are ran against the most recently built local Docker images, those are the images with the `:latest` tag which are also taken when no tag is specified. The `example` test suite runs against the remote Docker Images specified in the configuration in the `/deploy` directory.
+
```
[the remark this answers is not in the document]
rti: good catch. fixed

README.md
```diff
@@ -1,18 +1,21 @@
...
+Wikibase Suite (WBS) is a containerised, production ready [Wikibase](https://wikiba.se) system, that allows you to self host a knowledge graph similar to Wikidata.
 
-The wikibase release pipeline contains scripts used for building, testing and publishing Wikibase docker images.
+> 🔧 This document is for people developing WBS. If you want to host your own Wikibase instance, head over to the [WBS Deployment Kit documentation](./deploy/README.md). If you are looking for individual WBS Service Images, head over to [hub.docker.com/u/wikibase](https://hub.docker.com/u/wikibase).
 
-It contains a set of build targets defined in the [Makefile](./Makefile) which can be executed in two different ways.
+## Overview
 
-When [building](docs/topics/pipeline.md), use the [build.sh](build.sh) script.
+This repository contains Wikibase Suite tools used for
```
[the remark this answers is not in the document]
rti: fixed

build/QuickStatements/README.md
```diff
@@ -38,7 +38,7 @@ In order to authorize QuickStatements against Wikibase via OAuth, this container
 
 Likewise, Wikibase needs to be able to access QuickStatements for the OAuth callback on a host-recognizable address, set using `WIKIBASE_PUBLIC_URL`.
 
-Note that Docker Engine doesn't provide such addresses, so you will likely need to set up a reverse proxy (such as nginx or haproxy) alongside either public DNS entries or a local DNS server using entries that route to these container. See the Wikibase Suite example configuration for more guidance on how to set that up.
+Note that Docker Engine doesn't provide such addresses, so you will likely need to set up a reverse proxy (such as nginx or haproxy) alongside either public DNS entries or a local DNS server using entries that route to these container. See the Wikibase Suite Deployment Kit for more guidance on how to set that up.
```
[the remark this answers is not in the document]
rti: The build/* READMEs are out of scope of this PR. I will move this comment to #684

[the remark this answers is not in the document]
rti: This change here should not be here. But thanks for the input. Will be handled in the next PR.

test/specs/repo/queryservice.ts
rti: Good point. Deployment Kit IS very long. Maybe we can have one more talk about the name. You are not the only one questioning the excessive length ^^

README.md
```diff
@@ -58,6 +62,113 @@ To take advantage of the git hooks we've included, you'll need to configure git
...
+
+You can run the tests in the docker container locally as they are ran in CI using `test.sh`.
+
+## Examples usage of `./test.sh`:
+
+```bash
+# See all`./test.sh` CLI options
+./test.sh --help
+
+# To run all test suites
```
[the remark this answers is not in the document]
rti: cool. fixed

deploy/template.env
rti: done

deploy/README.md
rti: I mean, we do have the logs at the moment (`wikibase` container `/var/log/mediawiki`), which are written into the containers and would be lost on `down`.

Though, I'd rather put the logs in a volume or send them to stdout only than having the container carry state intentionally .

rti: So for this PR I think we can just go forward with `down`

lorenjohnson: > I mean, we do have the logs at the moment (`wikibase` container `/var/log/mediawiki`), which are written into the containers and would be lost on `down`.
> 
> 
> 
> Though, I'd rather put the logs in a volume or send them to stdout only than having the container carry state intentionally .

Good point, and agreed. Clearing the nom-trivial accumulation of log files could actually be a reason for `down` on its own:)

Also agree that ideally the logs go to STDOUT if there is no volume, otherwise they go to files. 

However I did try getting them going to STDOUT and there was an issue with it due to how things work in MediaWiki (I think they essentially need to be bounced to a file with another process pushing them to STDOUT), and otherwise they are rather overwhelming to the otherwise (debatably) more fundamental Apache logs...

Note: Will copy a link you this to the related Phan ticket. Nothing to do here now 👍🏼

deploy/README.md
rti: Thanks fixed.

rti: @adeeri I now put a whole paragraph with an example how the updating works. I think this is what you suggested earlier. Let me know if this works for you https://github.com/wmde/wikibase-release-pipeline/tree/T362226-example-docs-2/deploy#updating-and-versioning

deploy/README.md
rti: done

deploy/README.md
rti: I rewrote the upgrade section and merged in the migration guide. https://github.com/wmde/wikibase-release-pipeline/blob/T362226-example-docs-2/deploy/README.md#major-upgrades

datn (dismissed):
I've pushed my changes to the branch.

deploy/README.md
rti: marking this as resolved for now. feel free to provide feedback and suggest/make changes at any time @adeeri

deploy/README.md
rti: followup https://phabricator.wikimedia.org/T367023

README.md
rti: followup https://phabricator.wikimedia.org/T367024

test/specs/repo/queryservice.ts
rti: follow up in https://phabricator.wikimedia.org/T364957

PR #686, opened 2024-06-06
Semantic versioning of components

adeeri:


variables.env
```diff
@@ -28,7 +29,11 @@ MEDIAWIKI_VERSION=1.41.1
 # as per https://github.com/wikimedia/wikidata-query-rdf/tags
 # and https://archiva.wikimedia.org/repository/releases/org/wikidata/query/rdf/service/
 # Compare to previous version: https://github.com/wikimedia/wikidata-query-rdf/compare/query-service-parent-0.3.135...query-service-parent-0.3.137
+# TODO: Add our own (semantic) version to hide this one?
 WDQS_VERSION=0.3.137
+# TODO: Decide on the independent versions to start from
+WDQS_FRONTED_VERSION=$WMDE_RELEASE_VERSION
```
lorenjohnson, line 35: Here is my suggestion for this release: 

* 1.0.0 for all images that are attributed to the updates to the 1.39 release, including "deployment kit" for that series
* 2.0.0 for all images that are attributed to the updates to the 1.41 release, including "deployment kit" for that series
* 3.0.0 for all images that are attributed to the new 1.42 release, including "deployment kit" for that series

variables.env
```diff
@@ -153,10 +159,17 @@ WIKIBASEMANIFEST_COMMIT=8c7cdae4d5a1e68b19fcca8eb5c2a7b6e3dc3e55
...
+
+# ##############################################################################
+# Quickstatements
+# ##############################################################################
+# Updated automatically by ./build.sh update_hashes
 # https://github.com/magnusmanske/quickstatements/commits/master
 QUICKSTATEMENTS_COMMIT=c4b2c6b086b319aa32dcdd7a323edf188faaa66d
 # https://bitbucket.org/magnusmanske/magnustools/commits/branch/master
 MAGNUSTOOLS_COMMIT=5b8cea412000072a2c8753023c11472a4ac11ef5
+QUICKSTATEMENTS_VERSION=$WMDE_RELEASE_VERSION
```
lorenjohnson, line 172: Even though they will all be the same for now, I'd go ahead and make all of these there literal version number (e.g. 1.0.0, etc) vs reusing the WMDE Version Variable, as they will diverge going forward.

variables.env
```diff
@@ -39,6 +44,7 @@ WDQS_VERSION=0.3.137
 # We cannot change this at the moment
 ELASTICSEARCH_VERSION=7.10.2
 ELASTICSEARCH_IMAGE_URL=docker.elastic.co/elasticsearch/elasticsearch:7.10.2
+# TODO: do we need ELASTICSEARCH_BUNDLE_VERSION so that the tag can express plugin changes?
```
lorenjohnson, line 47: I am not sure I got your question, but if I'm getting you-- yes, I think this image should be versioned like: 

1.0.0+es-7.10.2

I don't think this image currently will diverge between our 3 different Wikibase versions we're releasing (it hasn't yet yet that I know of), so it would just be that version and build for all 3 releases.

versions.inc.sh
```diff
@@ -12,45 +12,41 @@ if [ -f ./local.env ]; then
...
+    local tags=(
+        "$version"
+        "${version%.*}"
+        "${version%%.*}"
+        "example-${EXAMPLE_VERSION%%.*}"
+    )
+
+    # Extra tag for wikibase
+    if [[ "$image_name" == "wikibase" ]]; then
+        tags+=("${WMDE_RELEASE_VERSION}_mw-${MEDIAWIKI_VERSION}")
```
lorenjohnson, line 47: `_${MEDIAWIKI_VERSION}` or `+${MEDIAWIKI_VERSION}`. I think its `+` as the MediaWiki version is "extra build information" added to the version. I may be wrong...

variables.env
rti: Would you also have different major versions for all images, even though there might be no breaking change in between them? Such as WDQS based on 0.3.135 and 0.3.137?

lorenjohnson: My suggestion here is only for this release, for the initial versions of everything. Updates from there, on all components of the system, strictly follow smear. So if I'm understanding your questions correctly, my answer would be no. When there is a patch update to WDQS and we make a new WBS WDQS image release, it gets a patch semvar bump (e.g. 1.0.0 to 1.0.1). Does that make sense?

variables.env
```diff
@@ -7,7 +7,8 @@
 # Append -prerelease while working on releases.
 # Remove -prerelease when ready to merge release to the release branch.
 # Call it main on the main branch.
-WMDE_RELEASE_VERSION=main
+WMDE_RELEASE_VERSION=2.3.0
```
rti, line 10: I'd remove this variable completely in order to stress that there is not central version number anymore.

variables.env
```diff
@@ -7,7 +7,8 @@
...
+EXAMPLE_VERSION=4.1.0
```
rti, line 11:
```suggestion
DEPLOYMENT_KIT_VERSION=4.1.0
```

variables.env
lorenjohnson: Perhaps what you're pointing out is that there may be no material difference between the WDQS image for 1.39 and the WDQS image for 1.42. If that is the case, then in what I'm suggesting is that the WBS WDQS image is only build once and remains v1.0.0 regardless, and the patch update would be 1.0.1.

versions.inc.sh
rti, line 47:
```suggestion
        tags+=("${WIKIBASE_VERSION}_mw-${MEDIAWIKI_VERSION}")
```
in order to match the container name pattern

versions.inc.sh
```diff
@@ -12,45 +12,41 @@ if [ -f ./local.env ]; then
...
+    fi
```
rti, line 48: I think we need an extra tag for elasticsearch and wdqs as well.

variables.env
rti: Yes, this is what I had in my head. So we would ship WDQS 1.0.0 with the 1.39 based Deployment Kit and maybe 1.0.1 with the 1.42 based Deployment Kit, given that there is just a "patch" change in WDQS between them? 

Or would you - for the first releases - still give individual major versions numbers to the WDQS versions of the different Deployment Kit versions?

rti: > then in what I'm suggesting is that the WBS WDQS image is only build once

I think our build system is not there yet.

versions.inc.sh
lorenjohnson: Probably missing something, but I think the `_` is departing from semvar and that we should be using `+` in any/all of these cases?

rti:
Say, we have now two branches for two Deployment Kit (DK) versions.

`wbs-dk-1` (including 1.39 based wikibase) and `wbs-dk-2` (including 1.42 based wikibase). Both build the WDQS 1.0.0 Service Image. So at the moment we will have two pipeline runs. Both build and upload the WDQS image to dockerhub. One image will get tags (1.0.0, 1.0, 1, wbs-dk-1) and the other one will get (1.0.0, 1.0, 1, wbs-dk-2). So serveral tags are moved from image to image, even though it should actually be the very same image.

This probably falls in the area of

> then in what I'm suggesting is that the WBS WDQS image is only build once

Just putting this comment here as a flag and to hear what you think about it. My first guts is we could ignore it for the first round of releases and fix it in a soon to be worked on follow up task.

variables.env
adeeri: It's used as the wikibase version. I can rename the variable to reflect that.

variables.env
adeeri: Yes, I think you got it, but just for reference, the idea was to add a variable which versions all elastic search components together, essentially hiding the upstream versions, so that we have more control in expressing the type of changes.

versions.inc.sh
adeeri: I can change it to a `+`, but do you also mean to discard the `mw`?

versions.inc.sh
adeeri: Just for clarity, which tags exactly would you add to each of them?

versions.inc.sh
lorenjohnson: Ah, yeah that was an oversight, please do keep the `mw` :)

variables.env
rti: Yes, same paged. Thanks for clarification.

> but we should be fine as long as rebuilding the same version comes up with the same thing each time

Kind of, I still expect one weirdness here, see my comment https://github.com/wmde/wikibase-release-pipeline/pull/686#issuecomment-2160222671

versions.inc.sh
adeeri: It looks like we can't use `+` in the tags
```
Error parsing reference: "wikibase/elasticsearch:1.0.0+7.10.2" is not a valid repository/tag: invalid reference format
```

lorenjohnson: Exciting. Well, it can be safely replaced with a `_` as that has no semantics in semver. It would probably be a discussion, but also we could consider just not adding this extra build information to the tagging. 

My not-too-carefully-thought-through proposal: Keep the semver scheme intact in your script, and at the point of tagging call a function which replaces all `+`s with `_`s in the tag.

adeeri: I prefer to keep the tag consistent throughout the stack to avoid confusion. But I could change the `-` to a `_` so that we don't unintentionally imply that it's a pre-release.

lorenjohnson: Agreed.

rti (changes requested):
In our initial drafts for the tagging we also had a build number appended to the version to create one definitely unique tag. Should we add this in this PR too?

basically it could be just
```sh
echo "someTag_build$(date +%Y%M%d%H%M%S)"
```

variables.env
```diff
@@ -7,7 +7,7 @@
 # Append -prerelease while working on releases.
 # Remove -prerelease when ready to merge release to the release branch.
 # Call it main on the main branch.
-WMDE_RELEASE_VERSION=main
+WBS_DEPLOY_VERSION=3.0.0
```
rti, line 10: Please update the docs on top as well

variables.env
```diff
@@ -16,24 +16,33 @@ WMDE_RELEASE_VERSION=main
...
+
 # Does not follow laws of god or men.
 # Update to the latest version for major releases.
 # Reason about changes for patch versions.
 # as per https://github.com/wikimedia/wikidata-query-rdf/tags
 # and https://archiva.wikimedia.org/repository/releases/org/wikidata/query/rdf/service/
 # Compare to previous version: https://github.com/wikimedia/wikidata-query-rdf/compare/query-service-parent-0.3.135...query-service-parent-0.3.137
 WDQS_VERSION=0.3.137
+WBS_WDQS_FRONTED_VERSION=1.0.0
+WBS_WDQS_PROXY_VERSION=1.0.0
```
rti, line 37: # ##############################################################################
# Wikidata Query Service Proxy Version
# ##############################################################################
# Update to reflect changes

variables.env
```diff
@@ -16,24 +16,33 @@ WMDE_RELEASE_VERSION=main
...
+WBS_WDQS_VERSION=1.0.0
+
 # Does not follow laws of god or men.
 # Update to the latest version for major releases.
 # Reason about changes for patch versions.
 # as per https://github.com/wikimedia/wikidata-query-rdf/tags
 # and https://archiva.wikimedia.org/repository/releases/org/wikidata/query/rdf/service/
 # Compare to previous version: https://github.com/wikimedia/wikidata-query-rdf/compare/query-service-parent-0.3.135...query-service-parent-0.3.137
 WDQS_VERSION=0.3.137
+WBS_WDQS_FRONTED_VERSION=1.0.0
```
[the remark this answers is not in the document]
adeeri: I actually like having all the wdqs related versions in the same block. Do you think it's too cluttered?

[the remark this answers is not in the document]
rti: It actually creates three different images for different purposes based on different tools. They just all relate to the WDQS. So I think it deserves a bit of clutter.

variables.env
rti, line 1: As we introduce the `deploy-3` tag in this version, do you think we should add the :deploy-3 thing to all the images in `deploy/docker-compose.yml` in this PR too?

adeeri: This was not on my radar since this work started before the times of `deploy` :) This sounds a bit out of scope for this initial set of minimal changes. Do you think we can do it in a small follow-up?

rti: https://phabricator.wikimedia.org/T368036

lorenjohnson (approved):
I assume the commented out things are what you were referring to that get turned-on at merge or later. Otherwise, LGTM! Great work 🤘🏼

variables.env
```diff
@@ -1,13 +1,10 @@
...
 # ##############################################################################
-# Our version string
+# Deploy version
 # ##############################################################################
-# Used for docker image tags on GHCR and dockerhub.
-# Append -prerelease while working on releases.
-# Remove -prerelease when ready to merge release to the release branch.
-# Call it main on the main branch.
-WMDE_RELEASE_VERSION=main
+# Used for tagging compatible docker images on GHCR and dockerhub.
```
rti, line 6:
```suggestion
# The version of Wikibase Suite Deploy
```

.github/workflows/build_test_publish_release.yml
```diff
@@ -1,24 +1,27 @@
 name: 📦 Build Test Tag and Publish Release
 
 on:
-  pull_request:
-    types:
-      - closed
+  # pull_request:
```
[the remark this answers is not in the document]
adeeri: Yes

.github/workflows/build_test_publish_release.yml
```diff
@@ -90,7 +94,12 @@ jobs:
           for image in "${images[@]}"; do
             image_path="wikibase/${image}"
             url_run_id="ghcr.io/${{ github.repository_owner }}/${image_path}:dev-${{ github.run_id }}"
-            version_tag="$(image_version $image)"
-            docker tag "$url_run_id" "${image_path}:${version_tag}"
-            docker push "${image_path}:${version_tag}"
+            version="$(image_version $image)"
+            tags=$(version_tags "${image}")
+            for tag in $tags; do
+              # docker tag "$url_run_id" "${image_path}:${tag}"
```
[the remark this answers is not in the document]
adeeri: Yup

variables.env
rti: sorry, it is above already

rti (approved):

PR #687, opened 2024-06-07
Remove LocalSettings.override.php option from Wikibase image

lorenjohnson:


deploy/README.md
```diff
@@ -118,25 +118,9 @@ docker compose up --wait
...
 
-The absence of `LocalSettings.php` is the trigger for the Wikibase container to run the [MediaWiki installer script](https://www.mediawiki.org/wiki/Manual:Install.php). If you need to run the installer again, remove the generated `LocalSettings.php` file and restart your instance.
-
-#### `config/LocalSettings.override.php`
-This file is the location for your MediaWiki and Wikibase customizations. If you need to regenerate your `LocalSettings.php`, your `LocalSettings.override.php` is the safe place that is never rewritten, even when running the [MediaWiki installer script](https://www.mediawiki.org/wiki/Manual:Install.php).
-
-If you want to, for example, change your site name, just put the following in your `LocalSettings.override.php`.
-```php
-<?php
-$wgSitename = "My New Site Name";
```
rti, line 131: Can you see a place where this example could still live? I kinda liked it.

lorenjohnson: For sure, will bring it back into that section.

rti (approved):
I like that you completely removed it now and we do not have this hidden feature in the wikibase image as a leftover.

lorenjohnson:
> feature

Yes, there is always the `/var/www/html/LocalSettings.d` chain to map into if we need something like the override in the future. That is the hidden feature which remains :)

deploy/README.md
lorenjohnson: Actually, I'm going to leave it out for now. Now that `config/LocalSettings.php` is the prescribed place to make this change the user would actually be changing an existing entry for `$wgSiteName` for this vs overriding it. However, I can try and formulate another example in this section with a configurable that might be added to make things more concrete for the user.

PR #689, opened 2024-06-12
T366812: Add user support channel to deploy/README.md

darthmon:
Replacing the telegram group as user support channel since the engineering team has no direct access to it. Instead I added the wikibase suite bug report template to encourage users start conversations directly with us via phabricator.


datn (approved):
LGTM

lorenjohnson (approved):
LGTM! It might be ok to keep or add back a mention of the Telegram group also as a place to interface with the larger community, but I imagine simple is good for this first pass.

rti (approved):

rti:
@darthmon I think you can merge by bypassing branch protections here and ignore the required status. It expects tests to pass but tests do not run on doc only changes. This is a pipeline bug. Sorry for that 😊

PR #690, opened 2024-06-13
T367476: Enable REST API

lorenjohnson:
Pending decision: https://phabricator.wikimedia.org/T367476


rti (changes requested):
I like the idea. I'd suggest that we add at least a basic smoke test for the REST API.

lorenjohnson:
> I like the idea. I'd suggest that we add at least a basic smoke test for the REST API.

While I certainly agree in principle, I would like to hold off on that there and add it as part of the scope of https://phabricator.wikimedia.org/T364847 as--debatably--this test will not have a great home of its own at the moment. I also think the risk is minimal to release without the smoke test in this case. We could decide to not mention the feature in our release announcement though if we wanted to save the highlighting it until there was a smoke test?

PR #698, opened 2024-06-26
T296812: add faq item about volume storage location

rti:
https://phabricator.wikimedia.org/T296812

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -326,6 +326,66 @@ It is possible to migrate an existing Wikibase installation to WBS Deploy. The g
 
 As of this writing, we can offer no specific recommendations for VPS providers to host Wikibase Suite. The suite has been tested successfully on various providers; as long as the [minimum technical requirements](#hardware) are met, it should run as expected.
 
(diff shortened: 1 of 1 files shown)


lorenjohnson (commented):
I'm thinking this would be better as a more brief summary and a link/s to the related Docker documentation.  My biggest concern here is the scope of what we maintain in our docs, which we need to discuss and decide on as a team.

darthmon (approved):
It looks good to me!

I see Loren's point of view but I also think that how the text has been cooked is personalised enough for our users and this is what matters most as I see it. 

However, it is true that we could also add a link to the docker docs for users who want to go look for further information.

datn:
Sorry for my lateness, I was out sick part of last week. I agree with @lorenjohnson -- this is out of scope for our work. I recommend a single helpful link to the relevant portion of the Docker docs as a courtesy.

datn (changes requested):
See my comment in the PR -- I think this is out of scope for our docs and should be replaced with a helpful link to the Docker docs.

PR #703, opened 2024-07-01
T368255: fix wikibase entrypoint for 1.39

rti:


lorenjohnson (approved):
1) I slightly prefer to leave this in the deprecated form across all 3 releases for now, as I think it has minimal to no impact currently, and it would reduce the diff to minimum (none) between versions for now until we get a better handle on things... 

2) However, I am also ok with moving this forward.

PR #711, opened 2024-07-03
T369133: job runner auto restart

rti:
https://phabricator.wikimedia.org/T369133


lorenjohnson (approved):
Yes, let's do that. Good catch. I think this should go head and be part of the 3/2/1 release if it's not a hassle.

rti:
> Yes, let's do that. Good catch. I think this should go head and be part of the 3/2/1 release if it's not a hassle.

@lorenjohnson part of all three release branches now

PR #715, opened 2024-07-04
Merge main to deploy-1

lorenjohnson:
The idea here is to maintain some history. Don't squash merge. Then the actual releases branches that get merged next will have the expect "just the hashes" and notes diff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,47 +134,23 @@ For more information on testing, see the [README](./test/README.md).
 
 ## Release process
 
-WBS Deployment Kit and WBS Service Images are released using this repository. The process involves defining all upstream component versions to be used, building service images, testing all the images together and finally publishing them.
+WBS Deploy and WBS Images are released using this repository. The process involves updating all upstream component versions to be used, building images, testing all the images together and finally publishing them.
 
 ### Release checklist Phabricator template
 
 ```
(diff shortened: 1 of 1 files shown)


rti:
@lorenjohnson  I think if the goal it to get everything from `main` also into the release of `1.0.0`, we should merge `main` into `T368255-release-1.0.0`. Otherwise we have two PRs, this and #710 that would release `1.0.0`.

rti:
Sorry, did not read the description carefully before commenting. The "double release" issue remains in my opinion. To archive the "same history" thing, we could merge main to a release prep branch, and merge (not squash merge) the release prep branch to `deploy-X`.

lorenjohnson:
Yup, I gotcha. I will rebase these 3 PRs to their related "prep" branches. 

Looking forward I think we should perhaps be kicking-off the workflow only when we tag a specific version (e.g. `v1.0.0`) and publish only then, otherwise just merging to the related major version branches doesn't do anything. That pattern would mesh perfectly with what I see coming up once we sort out our monorepo/version scripting/build tools stuff...

lorenjohnson:
Ah, I see. The prep branches are basically already up to date with main, hence their purpose. I'll just remind in PR review there to make sure each of those is current with main before they are merged down to the deploy-x branches, and close these 3 PRs.

PR #725, opened 2024-07-15
docs(wdqs): fix updater reinit help

rti:


rti:
@lorenjohnson thanks a lot

PR #731, opened 2024-07-17
T368373 (1): Docker Compose-based builds

lorenjohnson:
_A modest precursor to https://github.com/wmde/wikibase-release-pipeline/pull/733, but should be good to go forward on its own._

This configuration for builds reduces our boilerplate build code and I think is a little bit more clear then what we had, but the primary reason for this change is that it opens the door to a few things that are upcoming:

* Multi-platform builds using buildx can be readily turned-on
* I think this docker compose file can be used directly on DockerHub for auto builds (including multi-platform), if we end-up wanting that
* Publishing to Docker Hub becomes as simple as `docker compose push`. I think our adhoc Github CI code for pushing/publishing can be reduced to just that effectively further reducing our reliance on GitHub Actions specifics.
* This configuration for building a set of services I think is more standard and therefore more understandable to potential custom build users and/or contributors. 
* ...and some other things I'm not yet remembering, but will note up here as I do... :)

**Note:**

* Does away with the extra "build" / date image tag (to be discussed, but I think we should do away with it), otherwise does everything the current build script does. Running `./build.sh` in any case other than `./build.sh update_hashes` will forward all parameters to `docker compose build`, e.g. `./build.sh wikibase` becomes `docker compose ... build wikibase`.
* Still needs a couple of minor documentation updates to reflect the changes here


docker-compose.build.yml
```diff
@@ -0,0 +1,93 @@
+services:
+  wikibase:
+    image: ${WIKIBASE_SUITE_WIKIBASE_IMAGE_URL?}:${WBS_WIKIBASE_VERSION?}
```
lorenjohnson, line 3: The `?` tell yaml/compose to throw an error if the value doesn't exist. They are used throughout this file to avoid any accidental building of images which don't for some reason have all the env vars they need.

PR #732, opened 2024-07-17
T368373 (2): Skaffold-based builds POC (not CI integrated)

lorenjohnson:
Summary:

This works fine, but after testing this and the [Docker Compose based builds](https://github.com/wmde/wikibase-release-pipeline/pull/733), this compose build setup seemed a clear winner in terms of clarify and offered no less function for our use case. That Docker Compose build setup will easily translate to a Skaffold configuration if we ever want/need to go this direction.

### DIFF
--- a/run_skaffold.sh
+++ b/run_skaffold.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+set -o allexport
+# shellcheck disable=SC1091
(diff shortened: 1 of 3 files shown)


lorenjohnson:
@rti you are tagged for review on this just as an FYI. This won't move out of draft or be merged.

PR #733, opened 2024-07-17
T368373 (3): Docker Compose-based builds -- extended

lorenjohnson:
This and, https://github.com/wmde/wikibase-release-pipeline/pull/731 which it requires, is all putting us in a good place for independently updating and publishing Images (and Deploy). 

What is here also puts us in a place where switching to buildx-based multi-platform and Dockerhub auto builds both should be in easy reach. 

Generally with the more atomic / loosely couple structure here we may only need to add a semantic versioning library (Semantic Release, et al) and some accompany scripts for version and CHANGELOG generation, and end-up not needing any monorepo management tool.  TBD, but these changes I think make our build process better / more standard and clear:

* Moves Docker Compose based build setup for each image into each images respective build directory
* Moves build environment variables into `build/*/build.env` for each respective image, and creates a central `build/build.env` with shared values (e.g. `NODE_IMAGE_URL`, etc)
* Creates a central `build/docker-compose.build.yml` which references each of the individual docker compose files for each image for building all images at once
* Adds `build/builder/sh` which can run any `docker compose` command on top of the builder compose (e.g. `docker compose config` to see what will be built, etc)
* Changes `build.sh` in root to work as before but passes off to the `build/builder.sh` script, i.e. it defaults to `docker compose build` such that `./build.sh wikibase`, etc works as before
* Renames `build/*` directories to equal the name of the published docker images (i.e. Wikibase > wikibase, WDQS-frontend > wdqs-frontend, etc). Updates docs accordingly. 
* Renames all `WIKIBASE_SUITE_*` variables to `WBS_*` to match new version variables using this more concise standard.
* Removes dependency on `*_IMAGE_URL` env vars for image names and hardcodes `wikibase/wikibase` for image name in `docker-compose.override.yml` files for Client configuration for tests. This has already been done throughout the test suite and these had been missed. _We may actually want to go back to using the env vars from the build config, but for now just making this consistent._
* Adds a `version` file to `deploy` for tracking `WBS_DEPLOY_VERSION` (e.g. `WBS_DEPLOY_VERSION=3.0.0`)
* Does away with custom tagging script and instead uses env vars for the custom values (e.g. `WBS_WIKIBASE_VERSION_MINOR=3.0` and `WBS_WIKIBASE_VERSION_MAJOR=3`) to accommodate the compose setup without extra bash'ing to generate these values. I think it is clear enough. 

And, are we ok with these changes?:

* Does away with WBS Image version prefix on upstream version tags for images (e.g. `3.0.0-wdqs0.3.142` becomes simply `wdqs0.3.142`, etc). I think this is more clear, and more useful. Happy to bring back the version prefix if someone can convince me why its better :) 
* Does away with the special `:build<DATETIMESTAMP>` image tag for now. I realized I'm not a fan of this tag, as I think we should be paying attention to the build hash of the image itself for uniqueness / immutability and this extra metadata could be distracting and force updates where otherwise unnecessary. Also happy to bring it back if it is justified for some reason. 

TODO:

- [ ] Create/update `build/update_hashes.sh` to work with new repo structure


build/builder.sh
```diff
@@ -0,0 +1,25 @@
...
+IFS=$'\n\t'
+
+# Determine the directory of the script
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+
+set -o allexport
+# shellcheck disable=SC1091
+source "$SCRIPT_DIR/build.env"
+
+# Find and source all build.env files in the $SCRIPT_DIR/build/*/ directories
```
rti, line 12: Did you consider using docker compose `.env` file support? This could spare us this extra shell code here...


```
# .env
ENV_VAR=myvalue
```

```
# docker-compose.yml
services:
  build:
    build:
      context: .
      args:
        - BUILD_ENV_VAR=${ENV_VAR}
```

```
# Dockerfile
FROM debian
ARG BUILD_ENV_VAR
RUN echo "Using env var value: ${BUILD_ENV_VAR}"
```

Running:
```
❱ docker compose build     
[...]
=> CACHED [build 2/2] RUN echo "Using env var value: myvalue"
[...]
```

rti (commented):
Interesting concept... wrapping my head around this. Thanks for your work on this.

build/builder.sh
lorenjohnson: Ah yes, tried many things for that pre-eminently the `.env` file but also `services.<service_name>.env_file`, but none of them work to get env vars into build / Build Args in compose. I think that you'll see that it turns-out to be not be particularly important as you see the subsequent steps  as we once again move the parallel building outside of anything docker and use the nx task runner.

lorenjohnson: This branch is an entirely functional improvement of the current state of things, but also very much a rest stop on the way to the later work.

PR #735, opened 2024-07-18
T349803, T369642, T352252, T368382: nx monorepo setup for versioning, and CHANGELOG generation

lorenjohnson:
**Remaining work:**
- [x] Zoom-out and decide on our dev > release workflow including thinking through individual image releases (i.e. at what points to we publish releases, do we make it automatic from the start or phase into that, do we use Github Releases in addition to our CHANGELOGs, etc)
- [x] Introduce Actions way to manually kick-off a release for an individual product and restore Dockerhub push to work mostly as before
~- [ ] Update "Release checklist" / documentation accordingly~
- [ ] Merge, and make initial releases! 

---

Includes work detailed in: https://github.com/wmde/wikibase-release-pipeline/pull/733. 

**Addresses partially or completely all of these Phab tickets:**

* [T368382: Decision: Repository organisation (monorepo, etc)](https://phabricator.wikimedia.org/T368382) 
* [T349803: Automate versioning increment (major, minor, or patch version)](https://phabricator.wikimedia.org/T349803)
* [T352252: Design and implement CHANGELOGs](https://phabricator.wikimedia.org/T352252)
* [T369642: Make the release of individual WBS Images possible](https://phabricator.wikimedia.org/T369642)
* [T364924: Make ARM releases](https://phabricator.wikimedia.org/T364924)

**Implementation summary:** 

_Note: Initial work started in https://github.com/wmde/wikibase-release-pipeline/pull/731  and https://github.com/wmde/wikibase-release-pipeline/pull/733. These PRs can be re-opened, reviewed, and merged before  this PR if it is helpful for review._ 

Uses [nx](https://nx.dev/getting-started/intro) / [nx release](https://nx.dev/features/manage-releases) for managing all the steps of release up until actually pushing things to Docker Hub, including:

* Setting version (stored in `package.json` in each project)
* Generating `CHANGELOG.md` entries in each project from commits between previous version and the current

There are several more things we can do here, including integrating this into Github Actions, but this is the starting place which gives us what we need to manually generate a consistent and correct version of a project (either an image or deploy) independently along with an automatically generated `CHANGELOG.md` entries. 

While this uses the "non-Javascript" version of the nx installation to keep the noise down a bit, it does still require node and npm to be installed on the local dev machine. A Docker'ized version of this setup is probably possible, but may be adding a lot of extra friction (or not), TBD. For now I'm focused on getting the configuration correct.

_Note: The names of the mixed-case `build/*` directories have been made all lowercase to match the name of the image that goes up on Dockerhub. Because of how git handles case changes you will need to `rm -rf build` within the root of the repository before you will be able to successfully check out this branch. The same in reverse..._

**Adoption:**

1) Try it out!: `./nx release --first-release --dry-run` and look carefully at the log of the various things that would happen to understand how this all works.

2) Versioning will rely on our tags. Recommended: Removing all your local git tags and re-pull the current set of tags from the Github remote:

```
git tag -d $(git tag -l)
git fetch
```

3) Versions are automatically derived according to commits messages. To make sure this works correctly make commit messages using Conventional Commits prefixes (`feat:`, `fix:`, and `chore:`), and optionally but preferably with the product name scope (`feat(wikibase):`, etc). You could choose to forego this in development as we rely upon squash commits, but it is crucial that the squash commit message uses this format. Breaking Changes need to have "BREAKING CHANGE: <description>` in the commit footer message.  Please read through and get familiar with Conventional Commits here: https://www.conventionalcommits.org/en/v1.0.0. *

4) One engineer carefully make the first pre-release series using this sequence for each project, initially with `--dry-run`, but remove and go ahead and make the release once you're satisfied with the results:

```
./nx release version prerelease -p <project-name> --dry-run
./nx release changelog <version-returned-above> -p <project-name> --interactive --first-release --dry-run
```
For example, to see how the next pre-release of wikibase would be setup:

```
./nx release version prerelease -p wikibase --first-release --dry-run
./nx release changelog 3.0.1-0 -p wikibase --interactive --first-release --dry-run
```


lorenjohnson:
Bonus todos which I'll probably save for a follow-up ticket / PR:

- [ ] Rename nx-runner docker image to be wbs-builder (or similar) and integrate test dependencies and tasks into nx (result: one Dockerfile / container for build+test+release runner)
- [ ] Integrate linting as an nx task, including making more atomic and configuration more clear, etc

.prettierignore
```diff
@@ -0,0 +1,5 @@
+# Add files here to ignore them from prettier formatting
+/dist
+/coverage
```
rti, line 3: i think we should remove entries not relevant for us

Dockerfile
```diff
@@ -0,0 +1,37 @@
+# Use Node Iron LTS (20): https://nodejs.org/en/blog/release/v20.9.0
+FROM node:iron-bullseye-slim as wbs-dev-runner-base
```
rti, line 2:
```suggestion
FROM node:iron-bookworm-slim as wbs-dev-runner-base
```
bullseye is oldstable https://www.debian.org/releases/

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+        && pip3 install --no-cache-dir --upgrade \
+            pip \
+            setuptools \
+            requests \
+            bs4 \
+            lxml \
+            black
+# Install Docker CLI
+RUN curl -fsSL https://get.docker.com -o get-docker.sh | bash
+RUN sh get-docker.sh
```
rti, line 29:
```suggestion
# Install necessary packages
RUN apt-get update && \
    apt-get --no-install-recommends -y install \
        curl \
        docker.io \
        python3-requests \
        python3-bs4 \
        python3-lxml \
        black \
        && \
    rm -rf /var/lib/apt/lists/* && \
    ln -sf /usr/bin/python3 /usr/bin/python
```

rti: Install everything the debian way, do you remember why we had this get.docker.com thingy?

build/elasticsearch/build.env
```diff
@@ -0,0 +1,26 @@
+WBS_ELASTICSEARCH_VERSION=1.0.0
+WBS_ELASTICSEARCH_VERSION_MAJOR=1
+WBS_ELASTICSEARCH_VERSION_MINOR=1.0
```
rti, line 3: There is no way around this? Adee had this nice shell thingy, but we are not using this in shell context anymore, right?

build/elasticsearch/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+  elasticsearch:
+    image: ${WBS_ELASTICSEARCH_IMAGE_URL}:${WBS_ELASTICSEARCH_VERSION}
+    build:
+      context: .
+      args:
+        - ELASTICSEARCH_IMAGE_URL
+        - ELASTICSEARCH_PLUGIN_WIKIMEDIA_EXTRA
+        - ELASTICSEARCH_PLUGIN_WIKIMEDIA_HIGHLIGHTER
+      tags:
+        - ${WBS_ELASTICSEARCH_IMAGE_URL}:latest
```
rti, line 11: ATM we do not use `:latest` on docker hub. Would this end up on docker hub as well?

build/quickstatements/dockerhub.md
```diff
@@ -1,4 +1,4 @@
-# Wikibase Suite QuickStatements Image
+# Wikibase Suite quickstatements Image
```
rti, line 1:
```suggestion
# Wikibase Suite QuickStatements Image
```
I think here we can keep camel case... As a product name in text.

build/wdqs-frontend/dockerhub.md
```diff
@@ -1,4 +1,4 @@
-# Wikibase Suite WDQS-Frontend Image
+# Wikibase Suite wdqs-frontend Image
```
rti, line 1:
```suggestion
# Wikibase Suite WDQS Frontend Image
```

build/wdqs-proxy/dockerhub.md
```diff
@@ -1,4 +1,4 @@
-# Wikibase Suite WDQS-Proxy Image
+# Wikibase Suite wdqs-proxy Image
```
rti, line 1:
```suggestion
# Wikibase Suite WDQS Proxy Image
```

build/wikibase/dockerhub.md
```diff
@@ -1,4 +1,4 @@
-# Wikibase Suite Wikibase Image
+# Wikibase Suite wikibase Image
```
rti, line 1:
```suggestion
# Wikibase Suite Wikibase Image
```
Product Name... This Is Fine With Capital Letter I'd Say.

deploy/version
```diff
@@ -0,0 +1,2 @@
+# Version of this Deploy configuration (for reference when updating, does not need to be changed)
+WBS_DEPLOY_VERSION=3.0.0
```
rti, line 2: What is the idea behind this?

nx
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
rti: nixify me

test/helpers/default-functions.ts
```diff
@@ -140,7 +140,7 @@ export function defaultFunctions(): void {
 			await browser.waitUntil(
 				async () => {
 					const commandTextArray = await Promise.all(
-						commands.map( async ( command ) => command.getText() )
+						await commands.map( async ( command ) => command.getText() )
```
rti, line 143: This is why bitbuttler DOES have a product. Such things in another PR would be great. But also fine with me as long as it works...

test/package.json
```diff
@@ -1,59 +1,57 @@
 {
-	"name": "wikibase-release-selenium-testing",
-	"version": "0.1.0",
-	"private": true,
+	"name": "test",
+	"version": "3.0.0",
```
rti, line 3: :)

test/suites/docker-compose.override.yml
```diff
@@ -40,5 +40,19 @@ services:
...
+      - 4444:4444
+      - 5900:5900
+      - 7900:7900
+    volumes:
+      - .:/usr/src/test
+    shm_size: 2g
+    environment:
+      SE_NODE_MAX_SESSIONS: ${MAX_INSTANCES}
+      SE_VNC_NO_PASSWORD: 1
+
```
rti, line 56: Can you help me understand the restructuring here?

test/suites/repo_client/docker-compose.override.yml
```diff
@@ -9,7 +9,7 @@ services:
       - ./repo-extra-install.sh:/extra-install.sh
 
   wikibase-client:
-    image: ${WIKIBASE_SUITE_WIKIBASE_IMAGE_URL}
+    image: wikibase/wikibase
```
rti, line 12: I am not against hardcoding this, but then we would remove all the `WBS_*_IMAGE_URL` things from the `build.env` files, right?

rti (changes requested):
🤯 
WON-DER-FUL!

test/helpers/default-functions.ts
lorenjohnson: In this case this change is here to make TS transpilation not fail after udpated dependencies. It actually doesn't work as is and is a WIP...

.github/workflows/_build_test.yml
```diff
@@ -30,7 +51,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - run: ./build.sh ${{ matrix.component }}
+      # - uses: ./.github/actions/pull-ghcr
+      #   with:
+      #     github_token: ${{ secrets.GITHUB_TOKEN }}
+
+      - run: ./nx build ${{ matrix.component }}
```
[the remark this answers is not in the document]
lorenjohnson: Because our tests have to run from a Docker container (to access the Docker network for the services under test), we can't really parallelise effectively here and using multiple GH Action machines is our friend for now.

test/package.json
lorenjohnson: Agreed, but actually the naming and versioning here and in the root are necessitates for nx and npm workspaces to do dependency magic more than something I yet care a lot about. However, I do think what we're being forced to do here will serve us sooner than later in some ways especially around test versions...

test/suites/docker-compose.override.yml
lorenjohnson: The `wbs-dev-runner` form the root `Dockerfile` / `docker-compose.yml` is now our general purpose  container as test-runner had become. It is what is always ran whenever we run any `./nx <lint|test|build|release>` commands (so pretty much everything). Things are structured such with `TestEnv`  that this set of services is what is booted before each test suite is ran, which is when we also need the browser available. I think it ultimately fits a bit better here than anywhere else remaining, but it also was a needed move so the browser service wasn't needing to be managed along with he dev-runner... If that makes sense. It may still shift around a bit, but this is a good spot for it for the moment.

test/suites/repo_client/docker-compose.override.yml
lorenjohnson: Maybe. There are a couple things going on here, but for now this was just me making something consistent within the test setup. They were mostly all hardcoded except these. This was from the earlier PR with a note about it I think...

variables.env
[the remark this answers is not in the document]
lorenjohnson: byebye! don't be surprised if a env file reappears here again, but if so it will be all devops related vars...

build/quickstatements/dockerhub.md
lorenjohnson: Agreed.

Dockerfile
lorenjohnson: The official Debian packages for docker.io and docker-compose:
```
root@bc5467d1d442:/workspace# docker compose --version
Docker version 20.10.24+dfsg1, build 297e128

root@bc5467d1d442:/workspace# docker-compose --version
docker-compose version 1.29.2, build unknown
```
I'm afraid these are too far back for our purposes.

nx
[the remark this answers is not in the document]
lorenjohnson: Done

build/elasticsearch/docker-compose.yml
lorenjohnson: Really good point. I can't remember why I added this, but I think I had to in the current setup for CI or something. Will loop back on that, I'd rather not have it here at all and don't think I'd need or want it on Dockerhub, and would like to avoid breaking out to manage tags in the Publish area if we don't need to...

lorenjohnson: Yeah, this is the reason I had to add it: https://github.com/wmde/wikibase-release-pipeline/blob/73869322a22dce35a436ee89ccde4384f6ef75cf/.github/actions/pull-ghcr/action.yml#L36 ... Hmmm...

.gitignore
```diff
@@ -13,3 +15,45 @@ test/log
 test/mwlog
 test/suites/**/results/
 
+# === Added by nx installation
+
+# compiled output
+dist
+tmp
+/out-tsc
```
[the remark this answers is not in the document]
lorenjohnson: Agreed. I'll leave this unresolved until I get around to it :)

.vscode/extensions.json
```diff
@@ -0,0 +1,3 @@
+{
```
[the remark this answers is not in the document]
lorenjohnson: Yes, but I think I'll remove the prettier recommend as it isn't that relevant for us.

deploy/version
lorenjohnson: Good catch. This was from an earlier branch but is now covered by the inclusion of a `package.json` file.

test/helpers/default-functions.ts
lorenjohnson: So looping back here... Given that the QuickStatements tests all pass it does seem that this thing that Typescript complained about and I changed accordingly was necessary. In the dependency updates either the `$$` return type changed in WDIO (which from looking through release notes I think they did), or the TS settings got more picky.  Either way it is doing the right thing now :)

.github/workflows/_build_test.yml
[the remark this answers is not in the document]
rti: I thought about the build step, not about the test step. I thought the build could use `run-many` before uploading the built images to GHCR instead of building each image on a separate machine.

nx
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -e
+
+[ -f "local.env" ] || touch local.env
+# Make sure the docker network exists
+docker network create wbs-dev > /dev/null 2>&1 || true
```
lorenjohnson, line 7: Move below dockerenv condition...

.github/workflows/_build_test.yml
```diff
@@ -30,19 +29,42 @@ jobs:
...
+      #
+      # --set default.platform=linux/amd64,linux/arm64
+      #
+      # Timeout-minutes also need to be increased
+      - run: |
+          export $(grep -v '^#' build.env | xargs)
+          # export IMAGE_REGISTRY=ghcr.io
+          # export IMAGE_NAMESPACE=${{ github.repository_owner }}
+          export PR_BUILD_TAG=ghcr.io/${{ github.repository_owner }}/wikibase/${{ matrix.component }}:dev-${{ github.run_id }}
+          docker buildx bake --set default.tags=$PR_BUILD_TAG --push
```
lorenjohnson, line 66: I intend to replace this with `nx build <image-name> --set ... --push` once I get a couple more things sorted on that API, but this is fine for now

.github/workflows/_build_test.yml
```diff
@@ -30,19 +29,42 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - run: ./build.sh ${{ matrix.component }}
+      - name: Set up QEMU
```
lorenjohnson, line 32: This is not necessary at the moment, but takes ~2 seconds to boot and leaves us in a place to test more with multi-platform. Or we comment it out or remove it, adding a note to the multi-platform doc...

lint.sh
```diff
@@ -1,56 +1,89 @@
...
+# Linting JS, YAML, Python scripts, and general whitespace
+if $fix; then
+  echo "ℹ️ Fixing linting issues which can be fixed automatically"
 
-# ℹ️ Linting Markdown **/*.md
-if $SHOULD_FIX
-then
-  $RUN_TEST_RUNNER_CMD "npm run lint:fix-md"
-fi
+  if $prettier; then
```
lorenjohnson, line 55: This brings back the thing Ricky had setup with running prettier then linting, but makes it optional. I think this is a good compromise. The result will always be something that passes our eslint rules, and with prettier the code will be in some places--well--prettier (which I like).

nx
lorenjohnson: Actually this needs to be created either way as the test-services also rely upon it already existing.

build/docker-compose.build-template.yml
```diff
@@ -0,0 +1,8 @@
+services:
+  default:
+    image: ${IMAGE_REGISTRY:+${IMAGE_REGISTRY}/}${IMAGE_NAMESPACE?}:/${IMAGE_NAME?}:${IMAGE_VERSION?}
+    build:
+      tags:
+        - ${IMAGE_REGISTRY:+${IMAGE_REGISTRY}/}${IMAGE_NAMESPACE?}/${IMAGE_NAME?}:latest
```
rti, line 6: We do not tag `:latest` on dockerhub at the moment. How can we tag it when building (in order to access the latest builds while testing) but not push `:latest` on publish?

lorenjohnson: Indeed. The current Github Actions `_build_test.yml` step overwrites these tags and just pushes the constrcuted `dev<gh-run-id>` tag as before. The part about actually publishing these to Dockerhub is actually WIP. At this point we have several options I kinda want to evaluate which one we choose as/when we finish settling decisions about how we want to run our release process as of this work... This file also is extraneous and I will be removing it, so will resolve the convo, but for sure by the time we're pushing to Dockerhub things will likely look different and we certainly won't push the `latest` tag. Promise.

rti:
`./nx build` reports 
```
WARNING: current commit information was not captured by the build: failed to read current commit information with git rev-parse --is-inside-work-tree
```

This fixes it:
```
diff --git a/Dockerfile b/Dockerfile
index 1c5d7b6..e0a697c 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,14 +4,13 @@ FROM node:20-bookworm-slim AS wbs-dev-runner-base
 # WBS tests use the Selenium Standalone image, so no need for the embedded Chromium
 ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true
 
-WORKDIR /workspace
-
 # hadolint global ignore=DL3059
 SHELL [ "/bin/bash", "-o", "pipefail", "-c" ]
 
 # Install necessary packages except Docker and PNPM
 RUN apt-get update && \
     apt-get --no-install-recommends -y install \
+        git \
         curl \
         python3-pip \
         python3-venv \
@@ -33,6 +32,9 @@ RUN curl -fsSL https://get.docker.com -o get-docker.sh && \
     bash get-docker.sh && \
     rm get-docker.sh
 
+WORKDIR /workspace
+RUN git config --global --add safe.directory /workspace
+
 # Install PNPM
 RUN curl -fsSL https://get.pnpm.io/install.sh | ENV="$HOME/.bashrc" SHELL="$(which bash)" PNPM_VERSION=9.6.0 bash -
 ENV PATH="/root/.local/share/pnpm:${PATH}"
```

rti:
gg

> Bonus todos which I'll probably save for a follow-up ticket / PR:
> 
>     * [ ]  Rename nx-runner docker image to be wbs-builder (or similar) and integrate test dependencies and tasks into nx (result: one Dockerfile / container for build+test+release runner)
> 
>     * [ ]  Integrate linting as an nx task, including making more atomic and configuration more clear, etc

This is all done, isn't it?

.github/workflows/_build_test.yml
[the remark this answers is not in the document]
rti: I think this is not resolved. I'd really love to see nx in action here, as this is one of the core elements of our pipeline.

.github/workflows/build_test_publish_release.yml
```diff
@@ -32,22 +32,10 @@ jobs:
...
-
-          if git tag "$tag_name"; then
-            git push --tags origin "$tag_name"
-          else
-            echo "*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***"
-            echo "Cannot tag ${tag_name}. Most probably this tag is already given to another commit."
-            echo "Make sure to update the VERSION variable in variables.env to publish a new deploy release."
-            echo "*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***"
-          fi
+          ./nx release --dry-run
```
rti, line 38: Before merge, I guess this should not be a dry run anymore.

.github/workflows/build_test_publish_release.yml
```diff
@@ -75,27 +63,4 @@ jobs:
...
-            image_path="wikibase/${image}"
-            url_run_id="ghcr.io/${{ github.repository_owner }}/${image_path}:dev-${{ github.run_id }}"
-            version="$(image_version $image)"
-            tags=$(version_tags "${image}")
-            for tag in $tags; do
-              docker tag "$url_run_id" "${image_path}:${tag}"
-              docker push "${image_path}:${tag}"
-            done
-          done
+          ./nx release publish --dry-run
```
rti, line 66: Do the real thing?

Dockerfile
rti: Ok, I see. Actually, debian bookworm not offering docker compose v2 is the real blocker I see here. 

I am still concerned about pulling in stuff from so many different places here...

build/elasticsearch/build.env
rti: We still have this set of variables. You mentioned something about that yesterday. Unfortunately I forgot about the way forward you mentioned.

build/elasticsearch/docker-compose.yml
rti: This is not really solved, is it?

test/package.json
rti: The smiley above was not meant as making fun of it, more like being amazed how the structure holds for all parts of this repo. I think it is just consistent and makes sense looking forward.

build/elasticsearch/docker-compose.yml
rti: You explained below. Sorry for the noise. I lost track in the parallel convos going on here. Closing again.

lorenjohnson: As mentioned in the other comment the tagging configuration in the docker-compose files isn't currently applied in the build in CI, but is used locally. Once we sort out our publishing workflow this will likely change. That part is a WIP.

.github/workflows/_build_test.yml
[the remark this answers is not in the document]
lorenjohnson: This step will go back to a nx command once we sort out the publishing workflow.

Dockerfile
lorenjohnson: It's a dev container. I welcome any specific amendment you have in mind, but it right now protects us from doing essentially the same thing locally. Yes, we need the latest Docker and that script does just add the apt sources and keys and pulls the latest from the Docker apt repo.

build/build.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
```
rti, line 1:
```suggestion
#!/usr/bin/env bash
```

.github/workflows/build_test_publish_release.yml
lorenjohnson: Now that things are setup for manually triggering the releases the Dry Run is now an interactive Publish workflow option (with a default of true)

build/elasticsearch/build.env
lorenjohnson: The version tags are no parsed in a similar way to before and part of the `build/build.sh` script which handles both building for test runs (both locally and on GitHub), and building and pushing to Dockerhub. The version tags are only applied when the build is ran from the `nx release publish` task...

.github/workflows/_build_test.yml
```diff
@@ -29,21 +32,14 @@ jobs:
...
-
-      - uses: ./.github/actions/push-ghcr
+  
+      - uses: ./.github/actions/setup-environment
         with:
-          docker_image: wikibase/${{ matrix.component }}
           github_token: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Build and push ${{ matrix.imageName }}
+        run: pnpm exec nx build ${{ matrix.imageName }} --push
```
rti, line 41: Where is this `--push` coming from? I cannot find who defines it where?

build/build.sh
```diff
@@ -0,0 +1,115 @@
...
+IMAGE_NAME=$(jq -r '.name' package.json)
+
+# publish to Dockerhub
+if [ "$PUBLISH" == true ]; then
+	IMAGE_REGISTRY=dockerhub.io
+	IMAGE_NAMESPACE=wikibase
+# build/test in CI
+elif [ "$GITHUB_ACTIONS" == true ]; then
+	IMAGE_REGISTRY=ghcr.io
+  IMAGE_NAMESPACE="${GITHUB_REPOSITORY_OWNER}/wikibase"
```
rti, line 81: indention broken

.github/workflows/_build_test.yml
```diff
@@ -89,8 +83,6 @@ jobs:
           path: test/suites/**/results
 
   test-success:
-    # Post-test CI step - to check all tests succeeded
-    # Github Branch Protection rules require this to pass to allow merging
```
rti, line 93: I think this comment still makes sense. Otherwise the "true" step might be confusing.

.gitignore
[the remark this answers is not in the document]
rti: Still here :)

.github/workflows/build_and_publish_release.yml
rti, line 1: ❤️

build/build.sh
```diff
@@ -0,0 +1,115 @@
...
+fi
+
+# Change to the directory for the specified project
+cd "$1" || { echo "Failed to change directory to $1"; exit 1; }
+
+# Remove the first argument, leaving the rest for docker buildx build
+shift
+
+DRY_RUN=false
+PUBLISH=$([[ "$NX_TASK_TARGET_TARGET" == "nx-release-publish" ]] && echo true || echo false)
```
rti, line 20: I think this is not really intuitive and the behavior too implicit.

build/build.sh
```diff
@@ -0,0 +1,115 @@
...
+		"dev-${GITHUB_RUN_ID}"
+	)
+# local build
+else
+	# When not tagging anything but the image name the "latest" tag is by default applied,
+	# making that explicit here:
+	TAGS+=(
+		"latest"
+	)
+fi
```
rti, line 61: I think I'd prefer to have explicit flags for that. Would make more clear what is happening from the outside.

build/publish.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
+
+# !!! 
+# Note: This script is currently not in use. Multi-platform builds cannot be stored locally, but only
+# in a Docker registry, so we currently add the appropriate tags and push to the registry at build
+# time in build.sh. 
+# !!!!
```
rti, line 7: If it is not used, can we remove the file?

Also: I still have difficulties bringing multi platform builds into this context. It makes the change very complex.

build/build.sh
```diff
@@ -0,0 +1,115 @@
...
+
+# get image specific tags
+# shellcheck disable=SC1090
+source "$BUILD_ENV_FILE"
+eval "$(declare -p IMAGE_TAGS)"
+TAGS+=(
+	"${IMAGE_TAGS[@]}"
+)
```
rti, line 69: I think this is only required in `$PUBLISH == true` case

.github/workflows/build_and_publish_release.yml
```diff
@@ -0,0 +1,66 @@
...
+            DRY_RUN_FLAG=""
+          fi
+
+          if [ "${{ inputs.image_name }}" == "All images with unreleased changes" ]; then
+            PROJECT_ARG=""
+          else
+            PROJECT_ARG="-p ${{ inputs.image_name }}"
+          fi
+
+          pnpm exec nx release publish $PROJECT_ARG $DRY_RUN_FLAG
```
rti, line 66: When this is doing an actual "release: publish to dockerhub" and it triggered manually, where does the git tagging happen then?

nx.json
```diff
@@ -0,0 +1,46 @@
...
+		},
+		"build": {
+			"command": "build/build.sh {projectName}"
+		},
+		"update-commits": {
+			"command": "build/update-commits.sh {projectName}"
+		},
+		"nx-release-publish": {
+			"command": "build/build.sh {projectName}"
+		}
```
rti, line 44: I find it confusing that `build` and `nx-release-publish` are actually the same commands.

build/wdqs-proxy/README.md
```diff
@@ -167,10 +164,9 @@ Hooking into the internal filesystem can extend the functionality of this image.
 | `/etc/nginx/conf.d/wdqs.template` | Template for the nginx config (substituted to `/etc/nginx/conf.d/default.conf` at runtime)                |
 | `/etc/nginx/conf.d/default.conf`  | nginx config. To override this you must also use a custom entrypoint to avoid the file being overwritten. |
 
-
 ## Source
 
-This image is built from this [Dockerfile](https://github.com/wmde/wikibase-release-pipeline/blob/main/build/WDQS/Dockerfile).
+This image is built from this [Dockerfile](https://github.com/wmde/wikibase-release-pipeline/blob/main/build/wdqs/Dockerfile).
```
rti, line 169:
```suggestion
This image is built from this [Dockerfile](https://github.com/wmde/wikibase-release-pipeline/blob/main/build/wdqs-proxy/Dockerfile).
```

build/wdqs/build.env
```diff
@@ -0,0 +1,19 @@
...
+# https://github.com/docker-library/docs/tree/master/eclipse-temurin/README.md#supported-tags-and-respective-dockerfile-links
+# No debian based image released by eclipse-temurin, let take ubuntu here.
+# As of March 2024, WDQS' startup scripts depend on the JRE 8 command line option PrintGCDateStamps.
+# https://phabricator.wikimedia.org/T316103
+JRE_IMAGE_URL=eclipse-temurin:8u412-b08-jre-jammy
+
+# https://hub.docker.com/_/debian
+DEBIAN_IMAGE_URL=debian:bookworm-slim
+
+IMAGE_TAGS="qs${WDQS_VERSION}"
```
rti, line 19:
```suggestion
IMAGE_TAGS="wdqs${WDQS_VERSION}"
```

build/wikibase/entrypoint.sh
```diff
@@ -91,12 +91,10 @@ else
     php /var/www/html/maintenance/update.php --quick
 
     if [ -f /default-extra-install.sh ]; then
-        # shellcheck disable=SC1091
```
rti, line 94: why is this going away?

rti (commented):
Thanks a lot for your awesome work on this. What a big lift. Looking good! See my comments above.

build/build.sh
lorenjohnson: Indeed, good catch!

build/publish.sh
lorenjohnson: Yes, will do. Kept this in for reference in this review mostly.

.github/workflows/build_and_publish_release.yml
lorenjohnson: For now the workflow we discussed has a developer running `./nx release` locally before running this actual publishing step manually on github. Happy to change this to an automatic trigger based upon git version tag matching, but for now the actual kicking off of the release itself (updating versions, changelogs, creating git tags, and pushing) is kicked off locally so we can check it all over and amend the changelog and such as we wish which would be awkward to facilitate in an Actions workflow. 

When I look at it again next week perhaps I'll change the Dockerhub publish action to trigger off the version tags... Do you remember why we decided against that when we discussed on Monday?

nx.json
lorenjohnson: Mostly agree, but what do you suggest? Could rename the script to `build-or-publish.sh`? As it is it does at least mirror what we're for now doing in each case `buildx build`.

build/wikibase/entrypoint.sh
lorenjohnson: Our spellcheck installation can now
find the context and so sees this script

.github/workflows/build_and_publish_release.yml
rti: > Do you remember why we decided against that when we discussed on Monday?

I do not remember deciding against that.

I remember this two workflows:
1. Manually triggered github actions workflow to create the release, git tag it, generate changelogs, etc
2. Git tag triggered github actions workflow running build, docker tag, publish to dockerhub on a release git tag

This way we can run step 1 fully in CI for the default case and need no setup on dev  machines, but can also run it locally when we want customization (such as changelog tuning).

nx.json
rti: Ideally I'd like to have it explicit, maybe via arguments, like:

```
		"build": {
			"command": "build/build.sh {projectName}"
		},
		"nx-release-publish": {
			"command": "build/build.sh --publish {projectName}"
		}
```

lorenjohnson: Ha! Totally will do that. It was in fact how it was in my original implementation, and I took it out as a way for insisting that publish only happened through nx managed command which is currently the trigger.

.github/workflows/build_and_publish_release.yml
lorenjohnson: Certainly that be added by simply changing the current workflow which does `./nx publish <imageName>` to do `./nx release -p <imageName>`, and then creating another workflow which triggers on the version tag/s to then build and push to Dockerhub. That would in fact give the option to run that command as a Workflow,
or locally, which sounds fine to me. I'll do that, and prefer at least that the Dockerhub publishing is triggered automatically in Actions. 

To be clear there is no "setup" required to run this locally though (just Docker), and the result will be the same.

build/build.sh
lorenjohnson: When docker builds and "loads" it will assume the latest tag if no other tags are appled. We can leave this case out entirely and still the right hings happens. I preferred it explict.

.github/workflows/_build_test.yml
lorenjohnson: Any extra args to the `nx build` command / `build/build.sh` get forwarded to `docker buildx build . <here>`: https://docs.docker.com/reference/cli/docker/buildx/build/#push

.github/workflows/_build_test.yml
```diff
@@ -19,7 +22,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
```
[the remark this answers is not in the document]
lorenjohnson: As described in the earlier review pass, running through Docker or NX parallel runs will only run in one instance of GitHub Actions, and we won't get the benefit of Github Action running in parallel. i.e. it will be a lot slower, so keeping this as it.

.github/workflows/build_publish_image_release.yml
```diff
@@ -0,0 +1,56 @@
...
+    steps:
+      - name: Extract imageName and confirm it is a valid semantic version tag
+        id: extract_tag
+        run: |
+          # Extract the tag name
+          TAG_NAME=$(echo ${{ github.ref }} | sed 's/refs\/tags\///')
+
+          # Use regex to extract the imageName before the @ and confirm the version is valid semver version tag
+          if [[ "$TAG_NAME" =~ ^([^@]+)@([0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?(\+[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?)$ ]]; then
+            echo "::set-output name=imageName::${BASH_REMATCH[1]}"
```
rti, line 30: Lets add some more comments here

.github/workflows/_build_test.yml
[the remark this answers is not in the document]
rti: Have you tested it? 

I could imagine it does not make a big difference, but I can be wrong. Standard Github runners [have 4 cores](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories). Most of our build tasks are not well parallelized and anecdotal I can tell that quite some time ™ is spent downloading things.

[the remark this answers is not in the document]
rti: Not that I want you to try it now or in the context of this PR, I just want to understand if this is something we might want to look into in the future.

build/wdqs-proxy/README.md
lorenjohnson: This is changed in what is currently up here.

.github/workflows/_build_test.yml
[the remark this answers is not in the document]
lorenjohnson: I do think it would be slower on a single machine even with 4 cores, and notable given that booting up the whole stack for each suite is not a trivial amount of resources. More fundamentally though I don't think we can parallel run test suites at this time due to port and proxy name conflicts of the services under test.

build/wdqs-proxy/README.md
rti: Sorry, my bad. I looked at the wrong file. It IS resolved. Thanks for doing that!

.github/workflows/build_publish_image_release.yml
lorenjohnson: Ok, clarifying note added here

.github/workflows/_build_test.yml
rti: Got it, thank you

.github/workflows/_build_test.yml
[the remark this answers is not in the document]
rti: Resolving it in the context of this PR.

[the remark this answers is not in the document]
rti: Thanks for your explanations.

.github/workflows/build_publish_image_release.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Extract imageName and confirm it is a valid semantic version tag
+        id: extract_tag
+        run: |
+          # Extract the tag name
+          TAG_NAME=$(echo ${{ github.ref }} | sed 's/refs\/tags\///')
+
+          # Use regex to extract the imageName before the @ and confirm the version is valid semver version tag
+          if [[ "$TAG_NAME" =~ ^([^@]+)@([0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?(\+[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?)$ ]]; then
+            # Sets and passes the imageName output from this step, which is used in the subsequent step
+            # e.g. `steps.extract_tag.outputs.imageName` gets set to "wikibase" matched from "wikibase@1.01")
```
rti, line 31:
```suggestion
            # e.g. `steps.extract_tag.outputs.imageName` gets set to "wikibase" matched from "wikibase@1.0.1")
```
I think this is a typo.

.github/workflows/build_publish_image_release.yml
rti: Thanks. Looking good, marked potential typo above.

.github/workflows/_build_test.yml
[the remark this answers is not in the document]
rti: I had to try it ;)

You are right, it is slower.
4:36 without matrix https://github.com/wmde/wikibase-release-pipeline/actions/runs/10681174241
VS
3:#6 with matrix https://github.com/wmde/wikibase-release-pipeline/actions/runs/10680911492

.github/workflows/build_publish_image_release.yml
lorenjohnson: Fixed typo

README.md
```diff
@@ -131,7 +125,6 @@ MW_SCRIPT_PATH=/w
 
 For more information on testing, see the [README](./test/README.md).
 
-
 ## Release process
```
rti, line 128: TODO

.github/workflows/_build_test.yml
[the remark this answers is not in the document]
lorenjohnson: Ah, right. Sorry, I had lost track of the fact that we were talking about builds vs test. I’m fine with flattening this for now and eating the extra minute on each run. I don’t strongly prefer it but simplicity is nice, however I’d worry if we go towards turning-on multi-platform build with this setup that that little bit of extra might get compounded, so would rather leave it be for now.

nx.json
```diff
@@ -0,0 +1,46 @@
...
+				"projects": [ "deploy" ]
+			}
+		},
+		"projectsRelationship": "independent",
+		"version": {
+			"conventionalCommits": true
+		},
+		"changelog": {
+			"projectChangelogs": {
+				"createRelease": "github"
```
rti, line 25: Is there a reason to do Github releases? I'd vote for not adding them as I see no real benefit in having them for now...

package.json
```diff
@@ -0,0 +1,50 @@
+{
+	"name": "wikibase-release-pipeline",
+	"version": "3.0.0",
+	"license": "MIT",
```
rti, line 4: we are actually using BSD 3-Clause https://github.com/wmde/wikibase-release-pipeline/blob/main/LICENSE

lorenjohnson: Updated

README.md
rti: moved to https://phabricator.wikimedia.org/T373877

package.json
```diff
@@ -0,0 +1,50 @@
+{
+	"name": "wikibase-release-pipeline",
+	"version": "3.0.0",
+  "license": "BSD-3-Clause",
```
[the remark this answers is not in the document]
rti: https://github.com/wmde/wikibase-release-pipeline/actions/runs/10685770237/job/29619387975?pr=735

build/elasticsearch/build.env
rti: 👍🏾

.github/workflows/_build_test.yml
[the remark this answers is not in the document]
rti: Yes, matrix is ~20% faster than single machine parallelization.

rti (approved):
Whoa! What a great piece of work. Thank you! Looking forward to play around with it as soon this is merged to `main`.

The last remaining question I have is about github releases (see above).

nx.json
lorenjohnson: I generally think those release pages will be a helpful asset to making release announcements, but don't feel strongly about it. I would prefer to keep this turned-on for now so we can review how they feel on our first releases this sprint. We can always remove the entry/entries and turn it off if we don't like it.

rti: Agree. This is not the decision to maintain github releases from now on. Let's test it and evaluate afterwards. 👍🏾

PR #743, opened 2024-07-31
Remove default Mediawiki extensions tests

adeeri:
Follow-up on https://github.com/wmde/wikibase-release-pipeline/pull/741.
Remove testing of extensions that are bundled with Mediawiki.


test/specs/repo/extensions/scribunto.ts
```diff
@@ -1,40 +0,0 @@
...
-
-describe( 'Scribunto', function () {
-	beforeEach( async function () {
-		await browser.skipIfExtensionNotPresent( this, 'Scribunto' );
-	} );
-
-	it( 'Should be able to execute lua module', async function () {
-		// eslint-disable-next-line security/detect-non-literal-fs-filename
-		const fileContents = await readFile(
-			new URL( 'bananas.lua', import.meta.url ),
```
rti, line 12: If we want to remove the lua test, we want to remove the `bananas.lua` as well.

test/specs/repo/extensions/scribunto.ts
rti, line 1: I think I would like to keep the scribunto test as it does not only test the extension itself, but also ensures that our container installs lua appropriately.

test/specs/repo/extensions/visual-editor.ts
```diff
@@ -1,36 +0,0 @@
-import page from '../../../helpers/pages/page.js';
-
-describe( 'VisualEditor', function () {
-	beforeEach( async function () {
-		await browser.skipIfExtensionNotPresent( this, 'VisualEditor' );
-	} );
-
```
rti, line 7: This should go away I think, we can assume VE.

test/specs/repo/extensions/visual-editor.ts
rti, line 1: I think I would keep this one simple VE test. It does not hurt.

test/specs/repo/extensions/nuke.ts
```diff
@@ -1,49 +0,0 @@
...
-		await $( '.mw-checkbox-none' ).click();
-		await $( 'input[value="Vandalism"]' ).click();
-		await $( 'input[type="submit"]' ).click();
-
-		await browser.acceptAlert();
-
-		await expect( $( 'li*=has been queued for deletion' ) );
-	} );
-
-	it( 'Should show that a Nuke\'d page doesn\'t exist', async function () {
```
rti, line 44: I think this also tests JobRunner processing, which is part of the setup we create. So this might be still worth testing. But curious what other people think about this.

test/suites/confirm_edit/LocalSettings.php
rti, line 1: I like that we can remove this extra suite now

lorenjohnson (approved):
Code looks good, and very ok with wherever this lands! Here are my thoughts on the keep-it-in/leave-it-out debate--but summary is I'm very ok however it lands for now:

Most basically I'm a fan of removing all the extension specific tests which are not extensions we package. If we keep some of those tests as they test other things we want to keep testing (e.g. the Lua installation, et al), I think it would be best to update the test descriptions accordingly and/or make a code comment that we're "keeping this for now for X reason".  

_BONUS Commentary: I'm feeling a zoomed-out look at our testing strategy coming up soon (like in weeks/months, not days), and if that is the case I think we can sort all of this then regardless of any choices made here in this PR. In my experience having a clear, focused, and maintainable test suite which is generally trusted and understood is of priority over coverage. Needless to say coverage is very important, but in practice I find it rarely beneficial to sacrifice any of those other things for it. This isn't a strong opinion, just clarifying where I'm coming from._

test/specs/repo/extensions/scribunto.ts
darthmon: you may have already done it - sorry for that spam in that case: should we not then rename the test specifically to what it's testing?

test/specs/repo/extensions/visual-editor.ts
darthmon: I don't know this so I am just asking for clarification: "it does not hurt" but is it needed @rti ?

test/specs/repo/extensions/nuke.ts
darthmon: same as above (you may have already changed it, sorry for the spam in that case): is the JobRunner processing being tested somewhere else or only here? 
1) if it's already being tested somewhere else I would vote to kick it this part too
2) otherwise, I would keep it but changing the name making it clear to any eye what is being tested

darthmon:
yaaaaay!! so cool seeing all that code go! thanks @adeeri !

test/specs/repo/extensions/scribunto.ts
adeeri: @darthmon how would you rename it? What the test is testing hasn't changed

lorenjohnson:
I think we should keep all the specs which decisively check something related to our image build which aren't otherwise checked elsewhere, and don't bother with the renaming or rescoping them in this PR.

Alternatively, with the Scribunto spec as example, we could move it to `repo/wikibase-image-dependencies.ts` as what it is confirming that the Lua we build into the container installed correctly. Rename the `description` to `Wikibase Image Dependencies`, and leave the names of the individual test case as they are as they are indeed saying what they test just fine.

Using this example, here is the deeper cut of what I see as the discussion and scope beyond this current task:

I would move this spec from `repo/extensions/scribunto.ts` to `wikibase/extensions/dependencies.ts`(as in this spec tests that the dependencies which the image needs and installs are working. Lua is one of them). So the `description` would also change from `Scribuntu` to `Dependencies`, and a new `wikibase` Test Suite would be created at `test/suites/wikibase` with (for now) the same setup as the repo test suite.  There are a couple of things going on here, but the intention is to start moving towards atomic testing of each image, even if the spec or test suite setup still requires all the services. We can pair that down later. As we are currently setting up to release each of the images independently of each other we also eventually presumably want to be able to run tests relevant to each individual image.

lorenjohnson:
As for the Nuke test / JobRunner testing... The JobRunner implicitly tested in several other specs, look for `browser.waitForJobs()` as an indicator, and regardless without the JobRunner functioning many test will fail beyond the Nuke extension.

test/specs/repo/extensions/visual-editor.ts
rti: Well, I have never seen VE breaking, but I also lack experience in the stack to confidently state: "VE will not break". So my personal guts prefers to have a simple test ensuring that VE is in a usable state.

lorenjohnson: Does this test tests something particular about our packaging of MediaWiki? That is what I'm generally looking for in keeping around any of this.

rti: Yeah, not really. We really come to the question "What is our testing strategy and what do we want to ensure with our test suite?" Probably really out of scope for this PR. I am also ok with removing VE tests.

rti (approved):
👍🏾

adeeri:
As discussed here and online, I'm merging this in its current state and will follow up by looking deeper into the discussions we started here, mainly about Scribunto and Visual Editor tests.

rti:
merging in main for CI pipeline workaround

PR #751, opened 2024-08-14
fix(deploy): only expose proxy ports on host

rti:
https://phabricator.wikimedia.org/T372455


lorenjohnson (approved):
LGTM, and glad to reduce any noise for the Deploy config always. 

_One note: In anticipation of our updated nx-released based releases I'd consider leaving off the version bump. It will be harmless but won't translate directly to the new configuration on merge anyway._

PR #752, opened 2024-08-14
T372458-self-hosted-runners

rti:
WIP PR to support self hosted runners.

See https://github.com/rti/github-nixos-runner on how set up those machines.


lorenjohnson:
Closing for now but keeping branch around. Can be re-opened at anytime.

PR #754, opened 2024-08-15
fix: permissions for composer.local.json

rti:
Depending on docker daemon user and github action checkout user, permissions on this file could differ, even breaking the build if not readable by the current user in the container build process. This patch ensures correct permissions.

Appeared as a problem on a self hosted github actions runner.


lorenjohnson (changes requested):
Maybe a comment in the Dockerfile to remind our future selves why this seemed necessary.

lorenjohnson (approved):

PR #755, opened 2024-08-20
feat: Adds optional healthcheck scripts into our images

lorenjohnson:
Takes all in-line healthchecks specifics and puts them in a `healthcheck.sh` for each image. Not applying by default in image (i.e. as a `HEALTHCHECK` in the Dockerfile) to keep things flexible, but gives the responsibility for how to healthcheck back to the image--which is what Docker recommends.  Also removing this specific from Deploy removes one more possible change vector in the Deploy configuration, keeping things easier to maintain and upgrade for users over time. Also has the benefit of making a standard healthcheck option available for the images for use cases outside of Deploy.

In a related move this adds a new `IS_JOBRUNNER` option the Wikibase image instead of changing the command/entrypoint script. This allows the same health check script to be pointed which now will also healthcheck jobrunner when used in that mode. 

_This PR is setup with atomic conventional commits and would be best to do a regular merge vs squash._


rti (changes requested):
Sorry for the late review. Nice effort. I like the cleanup and agree with the motivation.

Due to the long wait, merge conflicts appeared. Do you mind fixing them?

lorenjohnson:
> Due to the long way, merge conflicts appeared. Do you mind fixing them?

Got it, re-applied on the current main.

build/wikibase/README.md
```diff
@@ -43,6 +43,7 @@ Variables in **bold** are required on first launch without `LocalSettings.php` i
 
 | Variable                     | Default    | Description                                                                                                                                                                                                  |
 | ---------------------------- | ---------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| **`IS_JOBRUNNER`**              | false  | If set to true the container will only run `maintenance/runJobs.php` instead of MediaWiki                                                                                                                                       |
```
rti, line 46: This is not a breaking change, is it? As I see it, the image can still be used with the old healthcheck style and be set to "job runner mode" by manually choosing ` /jobrunner-entrypoint.sh`. Is that correct?

build/wikibase/jobrunner-entrypoint.sh
rti, line 1: Confusing that Github puts "Empty file." here while it actually tries to say "no diff".

rti (approved):

build/elasticsearch/Dockerfile
```diff
@@ -12,6 +12,7 @@ RUN \
     org.wikimedia.search:extra:${ELASTICSEARCH_PLUGIN_WIKIMEDIA_EXTRA} && \
   ./bin/elasticsearch-plugin install \
     org.wikimedia.search.highlighter:experimental-highlighter-elasticsearch-plugin:${ELASTICSEARCH_PLUGIN_WIKIMEDIA_HIGHLIGHTER}
+COPY healthcheck.sh /healthcheck.sh
```
rti, line 15: Could you tell me about your thinking why not to use the `HEALTHCHECK` dockerfile instruction? What are the downsides?

build/wikibase/jobrunner-entrypoint.sh
lorenjohnson: Indeed. It is a file mode change from 644 top 755...

build/elasticsearch/Dockerfile
lorenjohnson: It is taking the lead from nginx, mariadb, etc images. They don't build in a HEALTHCHECK instruction in the their Dockerfiles but expect it to be manually applied via the CLI or a Compose file. One of the reasons for that in Kubernetes use of the image the HEALTHCHECK will be ignored, which can make it misleading to have it in the file. The other reason, as I understand it, is that using a HEALTHCHECK instructions results in a lot of extra logging that otherwise doesn't happen in the Compose scenario? Basically I think there is a case against baking in the HEALTHCHECK in the Dockerfile itself, but it's not a terribly strong case--just a slightly stronger case than the reasons to choose to have it there. https://www.reddit.com/r/kubernetes/comments/15rnhu8/comment/jwkerd7/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

build/wikibase/README.md
lorenjohnson: I'm pretty sure that is true, but I would think of this previous was as deprecated as of this update.

rti:
> This PR is setup with atomic conventional commits and would be best to do a regular merge vs squash.

Maybe we can now squash it again?

rti:
I like the idea to integrate the health checks in the images. I think this is the right place to put them

PR #758, opened 2024-09-02
fix(deploy): wikibase waits for elasticserch

rti:
On first lauch, wikibase needs to create indices in elasticsearch.

This patch models this dependency explicitly in order to prevent a race condition that lead to failures on Ubuntu 22.04 with docker-ce.

https://phabricator.wikimedia.org/T371162

### DIFF
--- a/deploy/docker-compose.yml
+++ b/deploy/docker-compose.yml
@@ -10,6 +10,8 @@ services:
     depends_on:
       mysql:
         condition: service_healthy
+      elasticsearch:
+        condition: service_healthy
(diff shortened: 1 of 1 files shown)


lorenjohnson (approved):
LGTM! Thanks for getting right on that one.

_Related side note, I would like us to consider removing the verbatim docker-compose.yml configs copied into the READMEs at some point though in exchange for just a link to the actual file in Deploy. Otherwise any such updates to Deploy will either mean we should update those as well or they are otherwise always a little out-of-date._

PR #759, opened 2024-09-03
docs(wdqs): details about the empty sync issue

rti:
This adds some documentations about the issue where WDQS crashes when restarted with an empty database that never synced entities from MediaWiki.

https://phabricator.wikimedia.org/T354266


datn (changes requested):
I pushed some small text changes to the branch

rti:
@datn Oh wow. Thank you so much. That definitely sounds way better.

Could you approve the PR in the current state? Or do you see anything work remaining?

datn (approved):

PR #762, opened 2024-09-04
docs(deploy): clarify running locally without dns

rti:
https://phabricator.wikimedia.org/T372177


datn:
Sorry, I was out sick. Looking at this now --

datn:
I made the language a little bit clearer, but I have a thought on the technical side of things as an old DNS head. "example.com" is indeed publicly resolvable, just not to the thing people might want it to resolve to :) 

```
$ host example.com
example.com has address 93.184.215.14
example.com has IPv6 address 2606:2800:21f:cb07:6820:80da:af6b:8b2c
example.com mail is handled by 0 .

$ whois 93.184.215.14
inetnum:        93.184.212.0 - 93.184.215.255
netname:        EDGECAST-NETBLK-03
descr:          NETBLK-03-EU-93-184-212-0-22
country:        EU
admin-c:        DS7892-RIPE
tech-c:         DS7892-RIPE
status:         ASSIGNED PA
mnt-by:         MNT-EDGECAST
created:        2012-06-22T21:48:10Z
last-modified:  2012-06-22T21:48:10Z
source:         RIPE # Filtered
% This query was served by the RIPE Database Query Service version 1.113.2 (DEXTER)
```

If we're going all the way out on a limb advising them to make /etc/hosts changes, we may want to suggest that they use either their own domain or a truly unresolvable domain like "yournamehere.wikibase". Just a thought --

rti:
So this is interesting. Even examples can be difficult 😄 

I like your point. [RFC2606](https://datatracker.ietf.org/doc/html/rfc2606#section-2) suggests to use e.g. `.example` as a TLD to prevent overlap with actual DNS entries. So I think we actually could switch completely to `*.example`.

rti:
@datn done. Do you mind re-reviewing?

datn (approved):
LGTM

PR #763, opened 2024-09-06
docs(deploy): do not use --wait with up

rti:
This removes the suggestion to use --wait with docker compose up. While this could be convenient, it hides the logs from the users. As a result, users tend to not post the relevant logs in tickets they create.


datn (approved):
By default our instructions should work the way we want them to. If users want to run in detached mode (which is available without --wait using the -d flag), they can read the docker instructions. I approve of this change.

PR #766, opened 2024-09-17
ci: Update GHA release workflow name, fixes  issue with publish not pushing to Dockerhub

lorenjohnson:


build/build.sh
```diff
@@ -55,6 +55,7 @@ if [ "$PUBLISH" == true ]; then
 	TAGS+=(
 		"${IMAGE_TAGS[@]}"
 	)
+	BUILD_ARGS+=("--push")
```
rti, line 58: was this broken before because this was missing?

.github/workflows/build_publish_image_release.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           if [[ "$TAG_NAME" =~ ^([^@]+)@([0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?(\+[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?)$ ]]; then
             # Sets and passes the imageName output from this step, which is used in the subsequent step
             # e.g. `steps.extract_tag.outputs.imageName` gets set to "wikibase" matched from "wikibase@1.0.1")
-            echo "::set-output name=imageName::${BASH_REMATCH[1]}"
+            echo "{imageName}=${BASH_REMATCH[1]}" >> $GITHUB_OUTPUT
```
rti, line 32: ❤️

build/build.sh
rti: maybe we can mention this fix in the PR title as this ends up in the `main` branch log

rti (approved):

build/build.sh
lorenjohnson: Yes, exactly. Tried a test release of an image and things worked as expected, except:

- It only built but didn't push the images, which this fixes
- There was an access error when the script tried to create the GH Release entry which is probably a simple fix, but figured we'd look at that when we get to testing this together

I'll update the title.

.github/workflows/build_publish_image_release.yml
lorenjohnson: Yeah still not the clearest API, but better.

Issue #385, opened 2022-11-23
How to change the host that is used for entities in WDQS? Default is `wikibase.svc`, I want it to be `localhost`


rdrg109:
# The context

I'm running Wikibase locally for experimentation purposes. In this section, I describe the installation process that I followed for getting it running.

I downloaded the files in the [`example`](https://github.com/wmde/wikibase-release-pipeline/tree/2f4394ce1f43dd8c179cc76874839d3e6fb10893/example) directory from the `main` branch as of the time of this writing [(permalink to commit)](https://github.com/wmde/wikibase-release-pipeline/tree/2f4394ce1f43dd8c179cc76874839d3e6fb10893). I didn't made a single modification to these files.

+ `example/docker-compose-extra.yml` [(permalink)](https://github.com/wmde/wikibase-release-pipeline/blob/2f4394ce1f43dd8c179cc76874839d3e6fb10893/example/docker-compose.extra.yml)
+ `example/docker-compose.yml` [(permalink)](https://github.com/wmde/wikibase-release-pipeline/blob/2f4394ce1f43dd8c179cc76874839d3e6fb10893/example/docker-compose.yml)
+ `example/extra-install.sh` [(permalink)](https://github.com/wmde/wikibase-release-pipeline/blob/2f4394ce1f43dd8c179cc76874839d3e6fb10893/example/extra-install.sh)
+ `example/jobrunner-entrypoint.sh` [(permalink)](https://github.com/wmde/wikibase-release-pipeline/blob/2f4394ce1f43dd8c179cc76874839d3e6fb10893/example/jobrunner-entrypoint.sh)
+ `example/template.env` [(permalink)](https://github.com/wmde/wikibase-release-pipeline/blob/2f4394ce1f43dd8c179cc76874839d3e6fb10893/example/template.env)

I then ran the `docker-compose` command as shown below.

```
$ docker-compose -f docker-compose.yml -f docker-compose.extra.yml up -d
```

```
WARN[0000] The "OAUTH_CONSUMER_KEY" variable is not set. Defaulting to a blank string. 
WARN[0000] The "OAUTH_CONSUMER_SECRET" variable is not set. Defaulting to a blank string. 
[+] Running 15/15
 ⠿ Network wikibase-docker_default                 Created  0.0s
 ⠿ Volume "wikibase-docker_query-service-data"     Created  0.0s
 ⠿ Volume "wikibase-docker_shared"                 Created  0.0s
[...]

PR #771, opened 2024-09-26
feat!: set wdqs concept URI

rti:
Allow setting the concept URI of entities in WDQS explicitly and distinct from the hostname used to contact wikibase.

This also fixes the problem WDQS not accepting entities when the wikibase host had a DNS name different from `wikibase` (which should be always the case, because `wikibase` is not a FQDN).

**Note: In deploy, items and properties in WDQS will now have the public wikibase hostname with https protocol as prefix. E.g. https://wikibase.example/**

Related:
- https://phabricator.wikimedia.org/T376016
- https://github.com/wmde/wikibase-release-pipeline/issues/385
- https://phabricator.wikimedia.org/T197682
- https://phabricator.wikimedia.org/T376106

BREAKING CHANGE: env var WIKIBASE_CONCEPT_URI now mandatory to run wdqs updater

### DIFF
--- a/build/wdqs/README.md
+++ b/build/wdqs/README.md
@@ -37,17 +37,18 @@ When running WDQS in a setup without WDQS-proxy, **please consider disabling the
 
 Variables in **bold** are required.
 
-| Variable                 | Default    | Description                                                                                                                                                                                                                                                               |
-| ------------------------ | ---------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| **`WIKIBASE_HOST`**      | "wikibase" | Hostname of the Wikibase service                                                                                                                                                                                                                                          |
-| **`WDQS_HOST`**          | "wdqs"     | WDQS hostname (this service)                                                                                                                                                                                                                                              |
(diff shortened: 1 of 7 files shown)


rti:
related https://phabricator.wikimedia.org/T197682

lorenjohnson (commented):
This looks like it should go forward, and would be nice to backport. However, given the major versioning to 1.x and 2.x issue we've been talking about I wonder if there is any sensible way to maintain the current way it functions (even if not optimal or broken) so that it can be a minor release? 🤷🏼‍♂️

rti:
How to handle this versioning, as discussed today with @lorenjohnson 

### WDQS bumps a major version
- because the interface changes the API in a breaking way
- it has a new required environment variable

### Deploy bumps a patch
- deploy bumps to WDQS 2 under the hood
- provides the required variable to WDQS
- for the user, the interface / API of deploy does not change

### TODO
- clean up commit messages to signify correct version bumps to deploy and WDQS projects
- do not squash merge, use merge commits

rti:
holding back for #782

lorenjohnson (approved):
LGTM!

PR #773, opened 2024-09-27
fix: drop privileges during docker build

rti:
Until now, the WDQS image started as root. In the entrypoint priveleges were dropped and we switched to the "blazegraph" user. With this patch, we switch to the "blazegraph" user already during build. So the container always uses this user and does not start as root anymore.

This fixes some issues when running an interactive bash inside the container and simplifies the updater reset workaround.

https://phabricator.wikimedia.org/T362963

### DIFF
--- a/build/wdqs/Dockerfile
+++ b/build/wdqs/Dockerfile
@@ -54,6 +54,7 @@ RUN apt-get update && \
 COPY --from=fetcher --chown=blazegraph:blazegraph /tmp/wdqs-service /wdqs
 
 RUN mkdir /var/log/wdqs && chown blazegraph /var/log/wdqs
+RUN mkdir /wdqs/data && chown blazegraph /wdqs/data
 
 # Don't set a memory limit otherwise bad things happen (OOMs)
 ENV MEMORY=""\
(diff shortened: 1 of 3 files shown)


lorenjohnson (approved):
This looks good! What will be the impact on existing use of the WDQS image I wonder? Would you anticipate any permissions issue?

rti:
> This looks good! What will be the impact on existing use of the WDQS image I wonder? Would you anticipate any permissions issue?

Well, before we had this thing to fix data access rights in the entrypoint in case somebody was booting the image with a data volume populated with an image that still used the old blazegraph user AKA `666`.

```sh
if [ -d /wdqs/data/ ]; then
  chown blazegraph:blazegraph -R /wdqs/data/
fi
```

We cannot do this anymore, because we are not root in the entrypoint with this change.

So if somebody would boot this new image with a data volume populated with an image before https://github.com/wmde/wikibase-release-pipeline/commit/9a775ccdf75415da97bd9e5c7632dafab3724c24 this could be a problem, because the access rights on the data do not match. But this change is already in [wmde.20](https://github.com/wmde/wikibase-release-pipeline/releases/tag/wmde.20), [wmde.18](https://github.com/wmde/wikibase-release-pipeline/releases/tag/wmde.18) and [wmde.17](https://github.com/wmde/wikibase-release-pipeline/releases/tag/wmde.17) (wmde.19 missed it somehow). So coming from releases from the "deploy-era" shouldn't be a problem.

What do you think @lorenjohnson ?

rti:
holding back for #782

PR #776, opened 2024-09-30
fix(build): always tag image versions

rti:
Before, local and ci builds would not tag image versions, just `latest`. This lead to e.g. `deploy` not picking up local builds. With this change, images get always tagged with versions numbers, regardless of where they are built.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -38,23 +38,24 @@ done
 
 # === Setup tags
 
+IMAGE_VERSION=$(jq -r '.version' package.json)
+IMAGE_VERSION_MAJOR=$(echo "$IMAGE_VERSION" | cut -d '.' -f 1)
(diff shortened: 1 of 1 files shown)


lorenjohnson (changes requested):
The reasoning behind not tagging the version until we have specifically decided to publish as a safeguard somewhat from different builds with the same version numbers floating around. I do still prefer it that way, but it does of course means we don't automatically use local builds when running deploy. I think we should look at how to solve that a different way and keep the version tagging as a publish event.

rti:
I see, thanks for your response. I understand that reasoning. I'd still like to noodle a bit on that topic, because I personally see pros in both approaches.

rti:
All right. I agree. Thanks for your thoughts.

> Having our workflow ensure that there is only one "authoritative" version under any version tag is worth prioritising I think, and central to why I see version tagging as a push/publish activity.

Closing this for now.

PR #781, opened 2024-10-02
chore(wikibase): bump mediawiki to 1.39.10, bump extensions

rti:


lorenjohnson:
Created a MediaWiki-1.39 branch off the head of main for now, and rebased this off that branch. That works fine, and I think is a good way to go about this.

_Interesting though, when I run `./nx release -p wikibase --dry-run` on this I get version 3.0.1, which is expected as we only have chores and fixes. We actually want to move to 1.1.0 as it's a MediaWiki patch and we decided to always move up minor versions as patch versions from MediaWiki aren't guaranteed to be just bug fixes, but include features sometimes... That aside, the release command also didn't catch the `chore(wikibase): ...` commit in the CHANGELOG and not sure immediately why. All the same for the other updated PR. Will explore more._

lorenjohnson:
@rti because MediaWiki updates are always considered to be Minor release due to potential new features we should conventional commit with "feat:" to get the right versioning automatically here. Going to re-write history here and on the other bump branch to reflect that.

rti:
@lorenjohnson I do remember us gravitating towards 
"patch mediawiki version bumps are minor versions" but I think it is not necessary. Rereading https://www.mediawiki.org/wiki/Version_lifecycle I now think we are pretty save calling it a patch release:

> [...] release branches continue to receive security updates for up to a year as of first release.

> Every point release will include updated i18n files as well as any bug fixes. No new features will be back-ported to point releases [...]

lorenjohnson:
> @lorenjohnson I do remember us gravitating towards "patch mediawiki version bumps are minor versions" but I think it is not necessary. Rereading https://www.mediawiki.org/wiki/Version_lifecycle I now think we are pretty save calling it a patch release:

Fine by me, with a small question of whether the current version also doesn't get any new features in the point releases?  Either way, shouldn't change this PR.

lorenjohnson (approved):

rti:
> Fine by me, with a small question of whether the current version also doesn't get any new features in the point releases? 

Yes, this is how I read it. From WMF perspective, all stable versions are not *current* anymore because they mainly care about `main` for production deployments.

>  Either way, shouldn't change this PR.

I think we should make sure that no commit mentions `feat:` then.

rti:
released with wikibase-lts@1.0.1 https://github.com/wmde/wikibase-release-pipeline/blob/5663695b2b1dad3cb9a0b64304274543d0d1a94c/build/wikibase-lts/CHANGELOG.md

PR #792, opened 2024-10-11
docs: include upstream docs in READMEs

rti:


lorenjohnson:
Is this a good place to mention that I think it is a mistake to include the docker-compose configurations in the READMEs of the individual images? The thought here is that it is a maintenance headache, or liability, as I think they are likely to get out-of-date and cause confusion. I think it would be better if we simply linked to the Deploy configuration as reference in those places in the READMEs, and leave out the docker-compose samples in the image READMEs. Should I make this its own ticket or can we make it part of the scope of what is being done here?

rti:
As we started to look a bit more into docs (e.g. #800) I would close this for now trusting that it will come up naturally again at some point soon.

PR #793, opened 2024-10-11
docs: fix update instructions

rti:
This ensures that docker compose always checks for new images when starting the WBS Deploy stack. Actually, we expected this behavior being present already. It is also already documented like that. That's why I consider this a fix and not a feat.

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -192,15 +192,16 @@ WBS Deploy may also receive minor and patch updates, but, as noted above, they a
 
 #### Minor and patch updates for WBS images
 
-Because WBS Deploy always references the latest minor and patch releases of compatible WBS images, non-breaking changes (including security updates) are applied automatically when re-creating Docker containers.
+Because WBS Deploy always references the latest minor and patch releases of compatible WBS images, non-breaking changes (including security updates) can be pulled at any time.
 
 This is always safe to do. Simply run:
 
 ```sh
 docker compose down
+docker compose pull
 docker compose up
 ```
-> 💡 In order to **prevent** new versions of WBS images being pulled on container restart, stop your containers using `docker compose stop` instead of `docker compose down`, which will keep the current containers intact. **Note**: this stops security updates from being applied. It is generally recommended to use `docker compose down`, which removes the containers and allows updates to be applied.
+> 💡 In order to automatically update images on every start, you can also use `docker compose up --pull always` to start your WBS Deploy stack.
 
 #### Minor and patch updates for WBS Deploy
 


lorenjohnson:
Hey @rti, I have a thought on this.  With `pull_policy: always` new images will be pulled even after just doing a `docker compose stop` and `docker compose up`, so on every ordinary reboot. So I think this `pull_policy` solution actually changes the behaviour we stated and makes patch and minor updates automatic, and no longer optional._ 

`docker compose pull` is another way to achieve this, such that the user determines explicitly when they want to receive updates and the images don't re-pull every time. I think this aligns with our current explanations in the README and I'd prefer adding it to the instructions at https://github.com/wmde/wikibase-release-pipeline/tree/main/deploy#minor-and-patch-updates-for-wbs-images, such that instead of:

```
docker compose down
docker compose up
```

We make the instructions there:

```
docker compose down
docker compose pull
docker compose up
```

And leave the service configuration as they are.  If we want to expand there README you could document this option to always pull and/or make it an env var, i.e. `pull_policy: ${PULL_POLICY}`, with a default of `missing` and a note about the option for fully automatic updates (on restart) by setting it it `always` (https://docs.docker.com/reference/compose-file/services/#pull_policy. 

Hope that makes sense.

lorenjohnson (changes requested):
Hey @rti, I have a thought on this.  With `pull_policy: always` new images will be pulled even after just doing a `docker compose stop` and `docker compose up`, so on every ordinary reboot. So I think this `pull_policy` solution actually changes the behaviour we stated and makes patch and minor updates automatic, and no longer optional.

`docker compose pull` is another way to achieve this, such that the user determines explicitly when they want to receive updates and the images don't re-pull every time. I think this aligns with our current explanations in the README and I'd prefer adding it to the instructions at https://github.com/wmde/wikibase-release-pipeline/tree/main/deploy#minor-and-patch-updates-for-wbs-images, such that instead of:

```
docker compose down
docker compose up
```

We make the instructions there:

```
docker compose down
docker compose pull
docker compose up
```

And leave the service configuration as they are.  If we want to expand there README you could document this option to always pull and/or make it an env var, i.e. `pull_policy: ${PULL_POLICY}`, with a default of `missing` and a note about the option for fully automatic updates (on restart) by setting it it `always` (https://docs.docker.com/reference/compose-file/services/#pull_policy. 

Hope that makes sense.

rti:
Thanks. Good catch. That does indeed make sense. Will update the PR accordingly.

deploy/README.md
```diff
@@ -192,15 +192,16 @@ WBS Deploy may also receive minor and patch updates, but, as noted above, they a
...
 #### Minor and patch updates for WBS images
 
-Because WBS Deploy always references the latest minor and patch releases of compatible WBS images, non-breaking changes (including security updates) are applied automatically when re-creating Docker containers.
+Because WBS Deploy always references the latest minor and patch releases of compatible WBS images, non-breaking changes (including security updates) can be pulled at any time.
 
 This is always safe to do. Simply run:
 
 ```sh
 docker compose down
+docker compose pull
```
lorenjohnson, line 201: `docker compose up --pull always`

For this section I think there isn't value to list both the down/up and down/"up --pull always" options as the section is exclusively about how to update.

lorenjohnson (changes requested):

deploy/README.md
rti: I do not agree. I think it is nice to have the pull as an explicit step here.

Your one liner is mentioned below as a hint too.

lorenjohnson: I see as unnecessary detail that can be distracting, but if you think it is better then I'm fine leaving it.

lorenjohnson (approved):
Please see final discussion really, and amend or merge as you see fit

deploy/README.md
rti: I'll go ahead and merge it now in order to have the actual fix available. The updating docs need some more general overhaul anyway as far as I am concerned. Let's consider the different options with the bigger context again.

rti:
Thanks lot for your thoughts and the review.

Issue #798, opened 2024-11-05
[WDQS] PREFIX not aligned with concept URI


fdeuze:
Setup:
Using WDQS:2 (wdqs0.3.142 from Oct 28, 2024) which solve correctly the issue to set the WIKIBASE_CONCEPT_URI required in wdqs-updater container (Note: However fix is needed in docker-compose.xml to set the correct version [here](https://github.com/wmde/wikibase-release-pipeline/blob/fe7f8918caf016d663a65827bd4efb7deda768f8/deploy/docker-compose.yml#L107C12-L107C27))

Issue:
The PREFIX in SPARQL query is not using the correct concept URI in wdqs container.
Indeed, setting the log "com.bigdata.rdf.sail.webapp.BigdataRDFContext" to level "debug", messages show:
PREFIX wdt: <http://wikibase/prop/direct/>
PREFIX wd: <http://wikibase/entity/>

instead of something like:

PREFIX wdt: <https://concept_uri/prop/direct/>
PREFIX wd: <https://concept_uri/entity/>

Recreation:
Any SPARQL query using prefix like wd or wdt from wdqs-fontend container UI.
For example this query doesn't return anything:
select *
where 
{
  ?s ?p wd:Q1
}
LIMIT 10

whereas this query return the expected result:
select *
where 
{
  ?s ?p <https://concept_uri/entity/Q1>
}
LIMIT 10

rti:
Hi,

Thanks a lot for reaching out!

> However fix is needed in docker-compose.xml to set the correct version [here](https://github.com/wmde/wikibase-release-pipeline/blob/fe7f8918caf016d663a65827bd4efb7deda768f8/deploy/docker-compose.yml#L107C12-L107C27)

Thanks for pointing that out. We released this as a fix to deploy yesterday https://github.com/wmde/wikibase-release-pipeline/releases/tag/deploy%403.0.3

We will have a look into the prefixes issue!

rti:
Closing this as the PR got merged to `main`

PR #804, opened 2024-11-08
fix: injected standard prefixes reference wikidata

rti:
~~As reported in #798, default prefixes injected by WDQS did not align with the `WIKIBASE_CONCEPT_URI` setting added in #771. This PR fixes it.~~

As reported in #798, the "standard prefixes" injected by WDQS do not align with the WIKIBASE_CONCEPT_URI setting introduced in #771, nor do they match the corresponding values for Wikidata. Currently, the WDQS-injected "standard prefixes" (such as `wd:`) reference `http://wikidata/`, which is the internal hostname of the wikibase/mediawiki container within the Docker network.

Particularly in the context of federation, it is essential to maintain the "standard prefixes" for referencing Wikidata [1] [2]. To accommodate the local Wikibase instance, prefixes can be set inline, as is currently done on wikibase.cloud [2]. This pull request addresses this discrepancy and updates the "standard prefixes" to point to Wikidata. Additional context can be found in [3].

[1] https://www.mediawiki.org/wiki/Wikibase/Wikibase.cloud/First_steps#View_your_data_using_the_Query_Service
[2] https://phabricator.wikimedia.org/T335448
[3] https://phabricator.wikimedia.org/T379232

### DIFF
--- a/build/wdqs/entrypoint.sh
+++ b/build/wdqs/entrypoint.sh
@@ -13,7 +13,7 @@ done
 
 set -eu
 
-export BLAZEGRAPH_OPTS="${BLAZEGRAPH_EXTRA_OPTS} -DwikibaseHost=${WIKIBASE_HOST}"
+export BLAZEGRAPH_OPTS="${BLAZEGRAPH_EXTRA_OPTS} -DwikibaseHost=${WIKIBASE_HOST} -DwikibaseConceptUri=${WIKIBASE_CONCEPT_URI}"
 export UPDATER_OPTS="-DwikibaseHost=${WIKIBASE_HOST} -DwikibaseMaxDaysBack=${WIKIBASE_MAX_DAYS_BACK}"
 
 envsubst < /templates/mwservices.json > /wdqs/mwservices.json
--- a/build/wdqs/entrypoint.sh
+++ b/build/wdqs/entrypoint.sh
@@ -13,7 +13,24 @@ done
 
 set -eu
 
-export BLAZEGRAPH_OPTS="${BLAZEGRAPH_EXTRA_OPTS} -DwikibaseHost=${WIKIBASE_HOST} -DwikibaseConceptUri=${WIKIBASE_CONCEPT_URI}"
+# Options provided to WDQS (blazegraph) when running as query service instance
+#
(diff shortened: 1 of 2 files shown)


darthmon (dismissed):
it all looks good to me :)

rti:
Thanks for your thoughts @tarrow, maybe this is not the best idea, even as a temporary solution. Thanks for catching that. Interestingly, the use of wikibaseHost goes back to 2021 to the birth of "example" (78f1acf3).

tarrow:
Yeah, TBH I think there was never a proper specification for how this should all behave.

I also wondered if the `WIKIBASE_HOST_PROPERTY` is important to be set for some other functionality (like the `label` service) or something.

I'd actually question if "magically injecting prefixes" was ever a desirable thing from a product standpoint. It saves a tiny bit of typing or clicking a "add prefixes" button in some UI at the expense of some real mystery for the user. 

I skimmed the SPARQL spec for how acceptable or not it would be to have this magic insertion of prefixes and couldn't find any strong evidence for or against.

tarrow (commented):
LGTM but I'm now also hoping that this doesn't impact in some way the "label" SERVICE; I wonder if you already checked this?

tarrow:
I was merely commenting and not requesting changes/approval) because I'd not care to tread on any toes of people not in my team :)

lorenjohnson (approved):
I haven't fully explored this, but to the 70% that I do grok what is happening here this seems like a clear and needed fix. The comment is appreciated. So this won't break current configuration for folks?

rti:
> LGTM but I'm now also hoping that this doesn't impact in some way the "label" SERVICE; I wonder if you already checked this?

Thanks for catching that. I just double checked and the label service works correctly in e.g.

```sparql
PREFIX PP: <https://wikibase.example/prop/>

SELECT ?aLabel
WHERE
{
  ?a PP:P1 ?c
  SERVICE wikibase:label { bd:serviceParam wikibase:language "[AUTO_LANGUAGE]". }
}
```

rti:
> So this won't break current configuration for folks?

Well, I would say: "it finally works as expected". So, "no", because I would consider this a bug fix.

PR #805, opened 2024-11-13
fix: workaround nx install error

rti:
On some machines, installing nx in the runner fails. Apparently, this is a known issue https://github.com/nrwl/nx/issues/27040. This PR adds the workaround mentioned.


lorenjohnson (approved):
👌🏼

PR #809, opened 2024-12-17
adding elasticsearch as a dependency

athalhammer:
Used `git checkout deploy-3` as described here: https://github.com/wmde/wikibase-release-pipeline/blob/main/deploy/README.md
and consistently ran into a race condition on `docker compose up` that had wikibase discovering that elasticsearch is not up yet (and therefore failing to initialize the indexes)

### DIFF
--- a/deploy/docker-compose.yml
+++ b/deploy/docker-compose.yml
@@ -10,6 +10,8 @@ services:
     depends_on:
       mysql:
         condition: service_healthy
+      elasticsearch:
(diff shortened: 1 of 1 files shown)


athalhammer:
I see that this is with tag `deploy@3.0.3`. So probably instead we should fix the readme, let me update the PR

athalhammer:
closing this in favor or #810

PR #810, opened 2024-12-17
updating readme to point to the latest version deploy tag

athalhammer:


lorenjohnson:
Hey @athalhammer, thank you for pointing this out. The intention is to maintain these version series tags vs updating to the latest tag in the README at each release. However, the catch here is that https://github.com/wmde/wikibase-release-pipeline/tree/refs/tags/deploy-3 was not pointing the latest version in the 3 series. Thank you for pointing this out and hope it didn't cause too much confusion.

PR #811, opened 2025-01-03
fix: set traefik restart policy to 'unless-stopped'

rti:
As reported in https://phabricator.wikimedia.org/T382200, traefik is the only service that remains stopped after a docker daemon restart (e.g. when rebooting the host). This change fixes traefik's restart policy to match the other services so that also traefik starts up automatically.


lorenjohnson (approved):
Good catch!

PR #813, opened 2025-01-06
chore: Add wikibase-lts and deploy-lts products

lorenjohnson:
Moves our Wikibase and Deploy LTS products beside our main releases of these products all in one repo. This reduces complexity, and solves a number of problems. The work in this branch completes much of what is needed to take this path, with the following remaining to be decided/addressed. 

- [ ] Decide if we really want to just tag lts-x.y.z on Dockerhub or use wikibase-lts as a new image (doesn't require any release configuration change)
- [ ] Update testing setup such the tests run on both LTS products as well as latest product releases.
  - [ ] Stretch goal: Run tests on both versions only when changes to the LTS products are detected, perhaps utilising `nx affected`.

Note: This branch includes a the yet unreleased Wikibase LTS 1.0.1 (MediaWiki 1.39.10 patch release), and the related Deploy LTS 1.0.1 releases already prepared including CHANGELOG updates (review those to see the format used).

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -71,8 +71,11 @@ else
 	)
 fi
 
-# transform TAGS to build args
 IMAGE_NAME=$(jq -r '.name' package.json)
+
+# Removes "-lts" from end of name if it exists.
+# LTS products are released on Dockerhub on the main image name
+IMAGE_NAME=${IMAGE_NAME%-lts}
 
 # publish to Dockerhub
 if [ "$PUBLISH" == true ]; then
--- a/build/wikibase-lts/CHANGELOG.md
+++ b/build/wikibase-lts/CHANGELOG.md
@@ -1,0 +1,18 @@
+## 3.0.2 (2024-10-28)
+
+
+### 🏡 Chore
+
+- bump mediawiki to 1.42.3, bump extensions
+
+## **wikibase@3.0.1** (2024-10-09)
+
+### 📖 Documentation
+
+- Link to MediaWiki bundled extensions
+- Switch from `.example.com` to `.example`
+
+### 🏡 Chore
+
+- Removes re-installation of extensions already packaged with MediaWiki (ConfirmEdit, Nuke, Scribunto, SyntaxHighlight_Geshi, VisualEditor)
+- Updates Wikibase EDTF extension
--- a/build/wikibase-lts/CHANGELOG.md
+++ b/build/wikibase-lts/CHANGELOG.md
@@ -1,11 +1,11 @@
-## 3.0.2 (2024-10-28)
+## 1.0.1 (2025-01-06)
 
 
 ### 🏡 Chore
 
 - bump mediawiki to 1.42.3, bump extensions
 
-## **wikibase@3.0.1** (2024-10-09)
+## (from **wikibase@3.0.1** (2024-10-09))
 
 ### 📖 Documentation
 
--- a/.github/workflows/build_publish_image_release.yml
+++ b/.github/workflows/build_publish_image_release.yml
@@ -9,6 +9,7 @@ on:
       - 'wdqs-frontend@*'
       - 'wdqs-proxy@*'
       - 'wikibase@*'
+      - 'wikibase-lts@*'
 
 permissions:
   contents: write
--- a/.github/workflows/create_release.yml
+++ b/.github/workflows/create_release.yml
@@ -10,12 +10,14 @@ on:
         default: 'wikibase'
         options:
           - deploy
+          - deploy-lts
           - elasticsearch
           - quickstatements
           - wdqs
           - wdqs-frontend
           - wdqs-proxy
           - wikibase
+          - wikibase-lts
           - All projects with unreleased changes
       dry_run:
         description: "Dry run, don't do it yet."
--- a/build/wikibase-lts/Dockerfile
+++ b/build/wikibase-lts/Dockerfile
@@ -1,0 +1,212 @@
+# Build configuration
+ARG PHP_IMAGE_URL
+ARG COMPOSER_IMAGE_URL
+
+# ###########################################################################
+# Based on https://github.com/wikimedia/mediawiki-docker/blob/1161796f04d6a6bcbec9fb4c67a8ce7248392403/1.41/apache/Dockerfile
+# hadolint ignore=DL3006
+FROM ${PHP_IMAGE_URL} as mediawiki
+
+SHELL ["/bin/bash", "-exu", "-c"]
+
+# System dependencies
+RUN apt-get update; \
+    apt-get install -y --no-install-recommends \
+        git \
+        gettext-base \
+        librsvg2-bin \
+        imagemagick \
+        # Required for SyntaxHighlighting
+        python3 \
+        # Required for Scribunto
+        lua5.1 \
+        ; \
+    rm -rf /var/lib/apt/lists/*
+
+# Install the PHP extensions we need
+# hadolint ignore=DL4006
+RUN savedAptMark="$(apt-mark showmanual)"; \
+    \
+    apt-get update; \
+    apt-get install -y --no-install-recommends \
+        libicu-dev \
+        libonig-dev \
+        libbz2-dev=1.* \
+        ; \
+    \
+    docker-php-ext-install -j "$(nproc)" \
+        bz2 \
+        calendar \
+        intl \
+        mbstring \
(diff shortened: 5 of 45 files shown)


lorenjohnson:
A couple of notes after looking into things a bit more today:

- After a bit of poking I have a pretty clear idea of how to amend the test setup to work well with this setup and only run LTS tests when there are changes in the LTS releases from the last release. 

- To highlight the perhaps obvious, if we were go this direction we can at any time inspect the diff between `build/wikibase` and `build/wikibase-lts` (to see what we might want or need to back port) we can run `git diff --no-index build/wikibase-lts build/wikibase` or a little more recklessly simply copy the contents over `build/wikibase` over `build/wikibase-lts` then do a `git diff`. All same of `deploy` and `deploy-its` of course. Git is our friend.

build/wikibase-lts/CHANGELOG.md
```diff
@@ -0,0 +1,18 @@
+## 1.0.1 (2025-01-06)
+
+
+### 🏡 Chore
+
+- bump mediawiki to 1.42.3, bump extensions
```
rti, line 6:
```suggestion
- bump mediawiki to 1.39.10, bump extensions
```

build/wikibase-lts/CHANGELOG.md
```diff
@@ -0,0 +1,18 @@
...
+
+## (from **wikibase@3.0.1** (2024-10-09))
```
rti, line 8: I am unsure whether we really want to include this information. Going forward, NX wouldn't generate it either. I think I would prefer just the list of changes since wikibase-lts@1.0.0

rti:
>  Decide if we really want to just tag lts-x.y.z on Dockerhub or use wikibase-lts as a new image (doesn't require any release configuration change)

I am still sympathizing with the idea to make wikibase and wikibase-lts one product on dockerhub. This is how the process could look like:

- Currently we maintain Wikibase 3.0.2 and Wikibase 1.0.0
- With this approach
  - Wikibase 3.0.2 (Mediawiki 1.42) would live in the `build/wikibase` directory as the wikibase "nx project", tagged with `wikibase@3.0.2` git tags
  - Wikibase 1.0.0 (Mediawiki 1.39 - LTS) would live in the `build/wikibase-lts` directory as the wikibase-lts "nx project", tagged with `wikibase-lts@1.0.0` git tags
  - Patching Wikibase 3 would happen in the `build/wikibase` directory and eventually lead to a release git tagged with `wikibase@3.0.3`
  - Patching Wikibase 1 would happen in the `build/wikibase-lts` directory and lead to a release git tagged with `wikibase-lts@1.0.1`
  - Now **What happens when we release a wikibase based on 1.43?** (anytime from now)
    - Mediawiki 1.39 [is still supported](https://www.mediawiki.org/wiki/Version_lifecycle), so we probably should support it too, so Wikibase 1 will continue to receive patches in the `wikibase-lts` directory
    - Mediawiki 1.43 will be released with Wikibase 4 from the `wikibase` directory.
    - Note that Mediawiki 1.43 is a LTS version...
  - Now **What happens when we release a wikibase based on 1.44?** (some time in summer) 
    - Mediawiki 1.44 will be released with Wikibase 5 from the `wikibase` directory.
    - Mediawiki 1.43 was an LTS version, so we can build our LTS version based on that too
    - We could copy the 1.43 based `build/wikibase` directory to `build/wikibase-lts` and continue maintaining it there, **continuing with version number 4, as wikibase-4.x.x**

This way, the switch from "released from `build/wikibase`" to "released from `build/wikibase-lts`" could be opaque to the user if we push from both "nx projects" to the same "dockerhub project" maintaining a consistent versioning throughout the full product lifecycle.

So users could onboard to Wikibase 4 now as soon as we release it and keep using it for years as long as we keep it in lts support. That we change the build folder name over time would not be relevant for the user.

I do agree though, that we need some special handling for the `wikibase-lts` git tag then. I wondered if we could have a new workflow triggered by `wikibase-lts@X.Y.Z` git tags. This workflow could just tag `wikibase@X.Y.Z` and push it to the repo to trigger the original publish workflow. Is this a hack or clever design? ;)

_This is raw thoughts dumped here, I might be missing something, or even a whole lot. The purpose of this comment is to further fuel the discussion._

lorenjohnson:
After more discussion the conclusion for now is to keep the wikibase naming for the product for both git tags and docker hub releases. This requires maybe only adding an exception here https://github.com/wmde/wikibase-release-pipeline/blob/1f8648f584db43f93cbf40d1fae02108c025ee76/build/build.sh#L75 for wikibase-lts to make the IMAGE_NAME wikibase

rti:
> After more discussion the conclusion for now is to keep the wikibase naming for the product for both git tags and docker hub releases.

I am not so sure. In my latest mental model, we only merge the LTS and non-LTS product only on dockerhub. But in the NX/git world, they would be two products.

build/build.sh
```diff
@@ -71,9 +71,12 @@ else
 	)
 fi
 
-# transform TAGS to build args
 IMAGE_NAME=$(jq -r '.name' package.json)
```
rti, line 74: Could we also just change the name in `package.json`? Without any code change?

rti: No, I think this is not a good idea. I think we need to keep the -lts around until we finally publish to dockerhub only. Otherwise we need to change too many things, e.g. local image versioning vs latest only, how to identify images when pushed to/pulled from GHCR...

lorenjohnson:
@rti following along with https://github.com/wmde/wikibase-release-pipeline/pull/815 👍🏼 

I was thinking about this all yesterday and doing a little research on how nx themselves manage their own LTS version. They use minor version branches such as 19.3.x. The more I rewound to looking at our options, the more I flip again to thinking we would be best to just have an lts branch. 

The reason I'm flip-flopping is realising that this current effort, once complete, will only solve one the issues we were having: correct generation of a version number for wikibase and deploy for lts. It does not solve or materially make better our story of how we track or reconcile differences or divergences, other than dependency versions, between latest and lts. And I actually think that directing users to a different branch than main for lts is as good if not a better user story than the separate directory idea.

As for our dev workflow it I think is pretty much the same either way, with slight differences either way:

**Updating our Wikibase LTS Image triggered by a MedaWiki LTS release: ** 

- Update `MEDIAWIKI_VERSION` in `build.env` and  then run `./nx update-commits wikibase` followed by `./nx release` and a commit / push to Github.

**Updating our Wikibae LTS Image with upstream changes from the latest WBS Wikibase release:**

- Patch files in the wikibase lts directory from files in the wikibase latest directory, commit, version, push/release. 

In the separate directories idea in this branch we have different options as to how we reconcile differences in the two branches scenario, but they aren't actually materially that different and one scenario I don't think has any huge advantage of the other... I could explain more if you wish on that. 

So, it seems to me that the two directory version of things here only solves one problem: `./nx release version` generating the expected version number as it will now be looking at a `wikibase-lts` tags to get to that new version number instead of looking to `wikibase` to get that version number and ending-up always trying to version as a next version of latest. 

I think I just need to look back into how to address that, we document very clearly how to make a latest release and an lts release, and automate the lts patch release once we get a coherent comms story that allows us to do that. 

As refreshing as something was about the "2 directories" story, after our long convo the other day I became convinced that LTS is not a separate product but simply an earlier version of the Wikibase product. Because it is only that. You helped me see that more clearly.

I think what we're doing now may actually end-up being just adding complexity where it needs not be there. I don't think it will actually notably simplify our workflow for releases or make anything possible that wasn't before with 2 branches.

rti:
@lorenjohnson I am not convinced. I still like the approach to only have one branch and I think it does make things simpler. It is a bit of a compromise, but the alternative to maintain more branches also does not come without compromises. We had it for a while. I would really like to try the single branch approach now and learn from it. 

One part why the single branch, lts-product approach still has quite some appeal for me is how  @RickiJay-WMDE put it: "This is a mono repo thing". It just makes sense how we set up the repo now. Branching the whole directory structure would be weird because we would have elasticsearch, quickstatements, all the tests, etc. in two branches, but they are actually always in sync. So this would be constant merging or cherry picking (as we had it). NX also gets hick ups, not only for version choosing but also for changelog generation, and depending on a patched version of nx could complicate things too.

rti:
@lorenjohnson Tests for LTS and latest are passing over here in #815. Would love to hear what you think about the actual changeset.

rti:
merged via #815

PR #825, opened 2025-02-03
docs: nx release and publish

rti:
## About

Add some documentation on how to use nx for releasing locally

## Status

## Dependencies

none


README.md
```diff
@@ -133,9 +133,37 @@ For more information on testing, see the [README](./test/README.md).
...
+# for all projects with unreleased changes
+./nx release --dry-run
+
+# Do a release (bump version, generate changelog and git tag it) for a single
+# project (e.g. wikibase).
+# Nothing will be pushed or published with this command, but a new commit with
+# changelog and package.json adjustments will be created locally. This should
+# happen  on `main`. Also, a new git tag with the new version of the released
+# project will be created locally.
+./nx release -p wikibase
```
lorenjohnson, line 151: I generally am a fan of orchestrating releases locally in this way, but how does this square with our current GHA setup for these things? Let's catch-up about it on today's meeting.

lorenjohnson (changes requested):
LGTM 👍  More or less I think this is the best way forward to make releases, but am wondering

README.md
rti: A release can be created just like this locally, or, in GHA using our [📦 Create a WBS Release](https://github.com/wmde/wikibase-release-pipeline/actions/workflows/create_release.yml) workflow.

Both paths will generate a release commit (ideally on `main`) and a tag. With the former you have to push from local to `origin/main` (works because we are all admins), the latter will push to `main` using our [Wikibase Suite Bot](https://github.com/wikibase-suite-bot).

As soon as the tag arrives, GHA will trigger [📦 Build and Publish a WBS Image Release to Dockerhub](https://github.com/wmde/wikibase-release-pipeline/actions/workflows/build_publish_image_release.yml) to publish to dockerhub in case of an image release.

lorenjohnson (approved):

PR #826, opened 2025-02-03
ci: fix all unreleased projects action

rti:
## About

Fixes a small bug in CI that broke the "release all unreleased" workflow

## Status

## Dependencies

none


lorenjohnson (approved):
It's been too long since I've been in here, but I assume this change was because this step doesn't actually tag ("release") deploy? If so this is fine, but might be nice if it read "All images with unreleased changes" but I may have missed something.

rti:
> It's been too long since I've been in here, but I assume this change was because this step doesn't actually tag ("release") deploy? If so this is fine, but might be nice if it read "All images with unreleased changes" but I may have missed something.

This works for all projects: images and deploy.

lorenjohnson:
> > It's been too long since I've been in here, but I assume this change was because this step doesn't actually tag ("release") deploy? If so this is fine, but might be nice if it read "All images with unreleased changes" but I may have missed something.
> 
> This works for all projects: images and deploy.

I see, so the clarifying verbiage felt unnecessary

PR #829, opened 2025-02-07
feat: allow configuration of wdqs frontend via `deploy/config`

rti:
--- 

**DO NOT SQUASH** - create merge commit

---

## About

Expose wdqs-frontend config file in `deploy/config` to make it easy to adjust page title, logo, example query source and similar.

**This is a breaking change for WDQS-frontend!**

## Status

## Dependencies

none


rti:
> Feedback from @lorenjohnson
> 
> > Keeping the env vars could make the change non-breaking, also, the env var interface might be useful in the future.

@lorenjohnson Thinking about it again I do like the removal of the env vars.

This PR moves the mechanisms to configure the image to "file based". It is debatable whether this is a good approach in general. But I think it is helpful to have one way to configure it. A hybrid env var / file based config interface might tend do be less clear, less intuitive and more complicated. WDYT?

lorenjohnson (approved):

PR #830, opened 2025-02-07
chore: update wdqs frontend host in template to match common setups

rti:
also covered in #833


deploy-lts/.env
rti, line 1: remove file

rti:
we will go with #833

PR #836, opened 2025-02-28
feat: user defined extensions

rti:
## About

Implements and documents a way to add user defined extensions via deploy.

## Status

ready to be reviewed, should be merged after #834 

## Dependencies

#834

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -206,6 +206,8 @@ docker compose up
 ```
 > 💡 In order to automatically update images on every start, you can also use `docker compose up --pull always` to start your WBS Deploy stack.
 
+If you installed user defined extensions in `config/extensions`, they might have updates too. Make sure to update them regularly too. See [User Defined Extension Docs](./config/extensions/README.md) for more information.
+
 #### Minor and patch updates for WBS Deploy
 
 WBS Deploy versions are tagged in git with tags such as `deploy@3.0.1`. Switching to a tag with the same major version will never trigger breaking changes. These updates are **always** considered safe. If you did not change `docker-compose.yml`, you can update simply by switching the git tag.
@@ -273,6 +275,10 @@ Read the [MediaWiki UPGRADE file](https://gerrit.wikimedia.org/r/plugins/gitiles
 No Wikibase-specific migrations are necessary.
 
 </p></details>
+
+##### Apply updates to user defined extension
+
+If you installed user defined extensions in `config/extensions`, they might require updates in order to be compatible with the new MediaWiki version. See [User Defined Extension Docs](./config/extensions/README.md) for more information.
 
 ##### Bring your instance back up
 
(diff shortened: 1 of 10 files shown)


test/suites/docker-compose.override.yml
```diff
@@ -10,7 +10,9 @@ services:
     image: wikibase/wikibase
     volumes:
       - wikibase-config:/config
+      - wikibase-extensions:/var/www/html/extensions/extensions
       - ./LocalSettings.override.php:/var/www/html/LocalSettings.d/LocalSettings.override.php
+      - ./empty.php:/var/www/html/LocalSettings.d/90_UserDefinedExtensions.php
```
lorenjohnson, line 15: Missing something there. Why this vs (or similar):

```
 - ./config/Extensions.php:/var/www/html/LocalSettings.d/90_UserDefinedExtensions.php
```

...and where is `empty.php`?

rti: `empty.php` is [here](https://github.com/wmde/wikibase-release-pipeline/blob/ead1b3ac0f786d9de19e31315d56451fe3cc1ae7/test/suites/empty.php#L1). This is the current working directory when running the tests.

I think I did not want to reference `../../deploy` in there because it felt fragile and like "crossing module borders".

rti:
@lorenjohnson thanks for your review and your thoughts.

> Is the image tolerant of the non-existence of 90_UserDefinedExtensions.php? If not can't we just create an empty 90_UserDefinedExtensions.php in the image that gets overwritten when the volume map happens?

Yes, it is. `90_UserDefinedExtensions.php` is just another `LocalSetttings.d` file loaded by [our LocalSettings loading mechanism](https://github.com/wmde/wikibase-release-pipeline/blob/ead1b3ac0f786d9de19e31315d56451fe3cc1ae7/build/wikibase/LocalSettings.wbs.php#L24) already present in the image.

> Is a simpler volume map API for custom extensions possible such that the user can volume map to /extensions and /extensions/Extensions.php vs these deeper image-specific (but debatably somewhat standard MediaWiki) paths?

It might be possible to bend the paths where MediaWiki is expecting extensions (e.g. [`$wgExtensionDirectory`](https://www.mediawiki.org/wiki/Manual:$wgExtensionDirectory)). I cannot see any benefits at the moment though and would argue that the config would become less intuitive and diverge more from common configs.

deploy/README.md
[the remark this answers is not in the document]
rti: done

test/suites/docker-compose.override.yml
lorenjohnson: From my viewpoint the coupling to the deploy configuration for both the user running all the services together, and as the basis for testing against is mostly already the case, desirable, and on purpose. Like if there is a divergence that causes the test setup to break, that is a good reason to update the test setup or tests? That said, I'm not terribly concerned about it as I see we're already bypassing any default "config" (understanding that there is currently nothing in the default config) from the deploy directory with a volume mapping and ok leaving it as is.

lorenjohnson (approved):
LGTM 👍  My one response is more important to me to share my perspective as basis for later convos if/when needed in this area, not so important to me that anything is changed here now.

PR #847, opened 2025-03-24
feat(ci): always test before publish: TEST DON'T SKIP CI RUN (DON'T MERGE)

lorenjohnson:


rti:
I wonder how this happened? Did you re-rerun the workflow manually two times?

image

lorenjohnson:
Yes, a couple things caused that and I am not sure I can perfectly remember what, but it was all due to manual intervention. The PR by default created against the `main` branch, and I didn't catch it until I'd created and so then cancelled the CI run and updated the base branch. There was also I think a sporadic failure of one test in the CI run on the earlier branch which this one is a duplicate of, so had to do a "Re-run failing jobs" on that one then re-force push this branch or something like that... To get a cleaner history we'd have to re-push both branch tests, but it shouldn't change the final result.

PR #862, opened 2025-04-17
feat: SELinux compatibility, configure Traefik via file

rti:
This pull request refactors the Traefik configuration to utilize a dynamic configuration file (`traefik-dynamic.yml`) instead of relying on Docker Compose labels.

The primary motivation for this change is to enhance compatibility with SELinux. SELinux, in its default enforcing mode, restricts container access to the Docker socket. Granting the Traefik container the necessary permissions to dynamically configure itself via labels requires running it in privileged mode, which is generally discouraged for security reasons.

By moving the Traefik routing and middleware configurations to a dedicated file, we eliminate the need for Traefik to interact directly with the Docker socket for its core configuration. This allows us to run the Traefik container in a non-privileged context even with , improving the overall security posture of the deployment when SELinux is enabled.

Additionally, this pull request introduces SELinux relabeling options (`:z`) to the volume mounts in the `docker-compose.yml` file. These labels ensure that the container processes have the correct permissions to access the mounted volumes. In essence, `:z` tells Docker to relabel the volume so that it can be shared between multiple containers. This is particularly important in SELinux environments to avoid permission denial errors.

This change ensures a more secure and SELinux-friendly deployment of the Wikibase stack on e.g. Fedora systems.


deploy/docker-compose.yml
```diff
@@ -226,6 +226,7 @@ services:
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock:ro
       - traefik-letsencrypt-data:/letsencrypt
+    privileged: true
```
rti, line 229: we should find a way around that

rti: related #860

RickiJay-WMDE (commented):
`./nx lint` works fine, but when I run `./nx build` on this branch I get permissions errors; when I run `./nx test` I get config errors; deploying results in `dependency failed to start: container wbs-deploy-wikibase-1 is unhealthy`. Happy to pair and trouble shoot any of those.

PR #865, opened 2025-05-05
feat: setup wikibase manifest external service mappings

rti:
We ship the [WikibaseManifest](https://www.mediawiki.org/wiki/Extension:WikibaseManifest) extension, but we did not set links to services we ship with WBS Deploy. This PR adds the URLs for WDQS GUI, WDQS sparql endpoint and Quickstatements.


addshore:
Happy to help figure out the local_rdf_namespaces if desired.
Is there a copy of WBSDeploy running in the cloud somewhere accessible to look at?

rti:
> Happy to help figure out the local_rdf_namespaces if desired. Is there a copy of WBSDeploy running in the cloud somewhere accessible to look at?

Hi @addshore, thanks for offering help here. This is definitely upcoming but out of scope of this PR and unfortunately even the couple of next sprints I guess.

At the moment, we do not have a demo instance running anywhere. This is a good idea though :)

rti:
@lorenjohnson thanks a lot for the review!

> Whether the wikibase image is being ran as a Job Runner or not is reliably indicated by the running of jobrunner-entrypoint.sh on start [...]

Great point, I will adjust that and set the env in the `jobrunner-entrypoint.sh`.

For 2 I see it differently. Deploy is the product that pieces everything together. That's why Deploy is also - IMHO - the right place to put the external services into the manifest. Other wikibase specific settings in the manifest might fit better into the image, I agree.

For 3 I am ambivalent, I did decide for the exception because Deploy is a defined set of services. I think this is the benefit of Deploy; it ships all those things and you can rely on them. I do hear you though, in terms of customization, it might be nice to be able to run without WDQS too. But then again, we do not officially support that as a configuration at the moment. So I would prefer to explicitly require WDQS.

lorenjohnson (approved):
LGTM! Comes out very tidy, I must say. I thought I might see the commiserate changes to Deploy also here, are you those going to go in a separate PR?

PR #867, opened 2025-05-20
Add operating system dependencies for Mediawiki extensions

olea:
See https://phabricator.wikimedia.org/T393666

Except for extension:3D for which I still don't have a running code.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,11 @@ RUN apt-get update && \
         git \
         curl \
         jq \
-        python3-pip \
+    	ghostscript imagemagick xpdf-utils \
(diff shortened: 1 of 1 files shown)


addshore:
It might be nice to include a comment saying which deps are required for which extension, so as to keep track of why things are included.
Perhaps a comment above each dep, saying why?

olea:
I'm not sure if dockerfile syntax allows it :-m


Ismael Olea

http://olea.org/diario/

olea:
I modified the patch to the proper place. Sorry.

rti (approved):

PR #869, opened 2025-06-12
feat(wikibase): Add operating system dependencies for Mediawiki extensions (WBS)

lorenjohnson:
The build is working, and our automated test suite is passing. I've also tested with manual installations of all extensions listed in the comments, and found them all to be working as expected with these dependency additions. 

- Removes a duplicate imagemagick dependency and updates comments slightly in the Wikibase Dockerfile from @olea's PR
- Some extensions included `package.json` files which by default are picked-up by our PNPM workspace configuration and causing build failures in those cases. So the `pnpm-workspace.yaml` file change sets PNPM to ignore any  `package.json` files found in `deploy/**` beyond the root `deploy/package.json`


rti (approved):
Thanks a lot @olea and @lorenjohnson for working on this. Looking good to me.

Also, thanks @lorenjohnson for catching the pnpm issue. This would now even allow adding these extensions within the wikibase build. Though, I would imaging the typical user flow being adding the extensions in [Deploy using the User Defined Extensions feature](https://github.com/wmde/wikibase-release-pipeline/tree/main/deploy#user-defined-extensions).

olea:
> Though, I would imaging the typical user flow being adding the extensions in [Deploy using the User Defined Extensions feature](https://github.com/wmde/wikibase-release-pipeline/tree/main/deploy#user-defined-extensions).

I still need to do a full configuration test for the extensions I want to use but, for the moment I think this approach works 👍

PR #870, opened 2025-06-12
feat: allow skip dispatched actions too

rti:


lorenjohnson (approved):
Fine with this of course, but without thinking about it too hard wondering why this was set as on before? I kinda remember but now forget :)

PR #874, opened 2025-07-03
chore!: bump mediawiki to 1.44.0

rti:


rti:
Solved, the resolution is an empty commit ;)

PR #875, opened 2025-07-07
docs: suggest METADATA_CALLBACK settings

rti:


deploy/template.env
```diff
@@ -24,9 +24,10 @@
 # If you change your mind you can always remove your instance from the list by
 # sending an E-Mail to wikibase-suite-support@wikimedia.de
 #
-# Set to true to opt-in to the list of known Wikibases.
-# Set to false to not take part.
-METADATA_CALLBACK=
+# Uncomment true to opt-in to the list of known Wikibases.
+# Uncomment false to not take part.
```
JeroenDeDauw, line 28: A clear indication of what the default behavior is would be nice. I saw the recent announce email and it also was not clear there, so I'm still wondering what the default is.

rti: Good point, the default is an error message asking you to decide.

lorenjohnson (approved):
I am approving this with the trivial note that the "this has to be set" logic is not yet true as of the time we're pushing this--and I'm ok with that.

PR #877, opened 2025-07-11
fix: do not require METADATA_CALLBACK for now

rti:


lorenjohnson (approved):
LGTM

FYI on a related note and for discussion elsewhere/at another time:

From the perspective of my current research on "1 click install" stuff it feels somewhat inevitably that we will introduce a configuration page or wizard of some sort as a first step after booting a VPS (via either a cloud config copy/paste or a one-liner like. https://wikiba.s/suite/install.sh). If I am correct in that, then these bits of UI currently embedded in comments in template.env and in this logic in docker-compose.yml could likely be eliminated. I do still see deploy/docker-compose.yml as a bit of a reference for any other configuration of the services, e.g. for k8n setups, on PAAS providers, or using services like fly.io. Currently our traefik setup I also think needs to be decoupled more to empower these cases, and I feel similarly about this inline entry point script.

PR #879, opened 2025-07-14
feat(deploy): T399695 - Initial Deploy config/setup wizard (Web and CLI)

lorenjohnson:
**To test feature see README and follow the instructions there:** 

https://github.com/wmde/wikibase-release-pipeline/tree/deploy-setup-script/deploy/setup/README.md

Related Phabricator ticket: [T399695](https://phabricator.wikimedia.org/T399695)

---

## Remaining TODOs

- [ ] Fix new issue with host names not showing up correctly in final message in `index.eta`
- [ ] Resolve review question about Debian/Ubuntu Docker repo source
- [ ] Look into deploy/setup/web-config.sh#41. The condition seems to be incorrect for when to notify the user they can close the terminal
- [ ] Look into clarifying `LAUNCH_ONLY` / re-run handling in launch.sh

---

- [ ] Re-run handling: Currently we basically are advising the user to start from scratch if they need to re-run, but in the case they do re-run the script:
  - [ ] Is the warning message when the repo is already found in ~/wbs adequately informative?
  - [ ] What do we do when we find an existing .env? Backup and replace it, or simply use it (currently we simply use it and the services proceed to try and launch)
- [ ] Double check Password validations to make sure they follow more tightly MediaWiki + MariaDB requirements,
      consider incorporating the common words dictionary lookup here. Have a "Generate" link for passwords or keep
      auto generated if ignored?
- [ ] Develop a Troubleshooting section in deploy/setup/README.md which informs user how to move past possible "invalid certificate" warnings on the setup wizard Web page, and on what to do if this same warning appears when the services are setup.
- [ ] Version or Not to Version
- [ ] Documentation diligence, refinement, and alignment from `deploy/README.md` through to `deploy/setup/README.md`
  - [ ] Review and refine Web Setup UI inline docs (including validation messages, host help, etc)
  - [ ] Submit for team review (Deepesha, Dan, et al)

---

## Testing notes

- Running `./setup.sh` directly from the repo’s `deploy/setup` directory skips Git install and repo clone.
To test those steps locally you have a few options:
  - Copy the script elsewhere and run it to trigger a fresh clone into ~/wbs/wikibase-release-pipeline
  - Use the curl command from the README (gets the pushed version from GitHub)
  - Temporarily comment out the “in repo” check in setup.sh
- Full setup process has been tested all the way to running services on Hetzner with the following distros: Ubuntu 22/24, Debian 11/12, Fedora 41/42, and CentOS 9/10
- Earlier testing was completed on DigitalOcean VPSs and I didn't find any differences with results on Hetzner, but it hasn't been tried again since substantial updates, nor has it been tried on other popular providers (e.g. Linode, et al)
- For now assuming official support for only the distros and versions mentioned (both for the setup script and in general for the Deploy product)

NOTE: The full debug log is available for each run at `/tmp/wbs-deploy-setup.log`

---

# Troubleshooting section notes

"Invalid Certifcate" warnings:

- Let's Encrypt rate limiting for *.nip.ip domains used for the setup installer means we may need to an "invalid cert" warning arising from self-signed SSL/TLS certificates when trying to access the page. This warning will also always appear when using --local option as all certificates in that scenario are self-signed, including as well for the WBS services.

- If for any reason you repeat the setup numerous times using the same values for Wikibase Public Host and/or Query Service Public Host, you may also see this invalid certificate warning for your site. If you want to continue to use those host names you will need to wait (how long?) before trying again so the certficate provider (Let's Encrypt) will unblock certificate provision for your designated hosts.
  
In the setup case it is safe and fully functional to bypass the warning and continue. See https://docs.vultr.com/how-to-bypass-the-https-warning-for-self-signed-ssl-tls-certificates for some guidance on how to do so.

---

# High level narrative introduction of the options available for setting up Wikibase Suite Deploy (DRAFT)

*Possibly a script for a short Wikibase Suite Deploy intro video?*

Wikibase Suite Deploy is the easiest way to get a full Wikibase environment up and running — all the moving parts integrated, configured, and ready to use.

There are two main ways to install it: the manual method, and the setup helper. Both get you to the same place: a fully working Wikibase Suite instance. The manual method is documented in the `deploy` directory’s README and walks you through each step yourself. The setup helper, found in the `deploy/setup` directory, does all of those steps for you — and more.

With the setup helper, you don’t even need to install Git or Docker yourself, or clone the repository. The script will handle that automatically if they aren’t already installed. It also provides:

- Default passwords for all required credentials
- Inline documentation as you enter your configuration
- A web-based interface so you can fill everything in easily
- Live status updates that tell you when your services are starting, when they’re ready, and if anything goes wrong

Once you’ve entered your configuration, the setup helper will launch your Wikibase Suite for the first time, let you know when it’s available, and tell you if there are any issues to fix.

The manual method works equally well — it’s just more hands-on and easier to make mistakes. It can also be useful for debugging if the setup helper isn’t working for some reason. But for most users, the setup helper is the recommended path.

At this stage, if something fails during installation — whether you’ve used the manual method or the setup helper — the safest way to recover is to delete your VPS instance, create a fresh one, and start again. That avoids tricky issues with partially completed installs or SSL certificate conflicts.

You can also run Wikibase Suite Deploy locally. This is only partially supported because it requires HTTPS for all services, and generating a valid SSL certificate for `localhost` isn’t straightforward. Locally, you’ll see a browser warning that the certificate is invalid — you can override this, and in most cases your browser won’t ask again. But QuickStatements won’t work locally at this time because it can’t authenticate without a valid certificate.

So:

- Use the setup helper for the easiest experience
- Use the manual method if you need fine-grained control or are troubleshooting

These instructions are all about getting to your first running instance. For day-to-day operations — things like backing up or resetting your data, starting and stopping services, or debugging in production — you’ll find more details in the main `deploy` README.

One last note: the configuration you enter during setup is for initial deployment only. It’s written to the `.env` file in the deploy directory. Changing these values after the fact won’t reconfigure your running instance — in some cases it will do nothing, in others it may cause unpredictable results.

If you get stuck, the Wikibase Suite team is here to help — reach out, and we’ll work with you to get your instance up and running.

---

- Adds a Deploy Setup script which is started with a one-liner curl command on a new VPS and complete configuration through a Web page or prompts on the command line

### Implemented

- User can easily setup WBS Deploy on a VPS provider of their choice
- User can configure the initial WBS Deploy setup through a easy to use web based UI with inline documentation
- User is clearly notified when WBS Deploy is done being configured and up and running
- User can trust the access information (passwords) for the WBS services are stored in a secure way (not in plain text in a .env file)
- User can create a WBS Deploy instance without having to register domain names
- User can configure the initial WBS Deploy setup through a easy to use text based CLI with inline documentation

### Won't be implemented

- User can setup WBS Deploy to run on localhost without external access and all services work as expected without warnings or errors
- User can easily and reliably use all WBS Deploy services in a localhost configuration
- User has a clear indication when a service doesn't boot or is left in an unhealthy status, and is provided instructions for debugging / how to proceed
- User can press a "Reset" button to stop all services, DESTROY ALL CURRENT DATA, change/re-apply configuration, and re-start services. 
  - docker compose down --volumes`
  - rm -rf config
  - git checkout config
- User can monitor docker compose service status in a WBS provided dashboard
- User can start/stop/reset docker compose service in a WBS provided dashboard


rti:
> User can trust the access information (passwords) for the WBS services are stored in a secure way (not in plain text in a .env file)

Unfortunately, MediaWiki still stores the DB password in plain text in the `LocalSettings.php` file. Definitely not in scope of this work, just noting to keep it on your radar.

rti:
> In Progress

Is not considered scope of this PR, correct?

deploy/setup/public/.keep
rti, line 1: Do we really need this file? The dir is nonempty anyway.

deploy/setup/package.json
```diff
@@ -0,0 +1,21 @@
+{
+	"name": "wbs-deploy-setup",
+	"version": "1.0.0",
```
rti, line 3: This is not integrated in nx, because it is not a separate product I suppose? Do we really track a version for that or can we maybe even just skip this field and consider this package part of deploy?

deploy/setup/scripts/cli-config.sh
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env bash
+set -e
+
+echo "🔧 Wikibase Deploy CLI Installer"
+
+# --- Parse args passed from go.sh ---
```
rti: Is it now called `launch.sh`?

deploy/setup/scripts/cli-config.sh
```diff
@@ -0,0 +1,81 @@
...
+VERBOSE="${VERBOSE:-false}"
+LOCALHOST="${LOCALHOST:-false}"
+
+# --- Setup logging ---
+# shellcheck disable=SC1091
+source "$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/_logging.sh"
+
+ENV_FILE_PATH="$DEPLOY_DIR/.env"
+PUBLIC_IP=$(curl --silent --show-error --fail https://api.ipify.org)
+
```
rti, line 18: Apparently logging is not used in this file

deploy/setup/scripts/web-config.sh
```diff
@@ -0,0 +1,101 @@
...
+VIEWS_DIR="$SETUP_DIR/views"
+
+# -- Setup logging --
+# shellcheck disable=SC1091
+source "$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/_logging.sh"
+
+generate_cert_for_setup_webserver() {
+  log "Generating TLS certificate for setup page..."
+
+  if [ "$LOCALHOST" = true ]; then
```
rti, line 27: Have you considered using Traefik for this? This could spare us the manual steps as well as use tech we already depend on anyway.

deploy/setup/scripts/web-config.sh
```diff
@@ -0,0 +1,101 @@
...
+  log_cmd "docker build $LOAD_FLAG -t wikibase/deploy-setup-webserver ."
+  log_cmd "docker run -d \
+    -p $SETUP_PORT:443 \
+    -v $DEPLOY_DIR:/app/deploy \
+    -v $VIEWS_DIR:/app/views \
+    -v $CERTS_DIR:/app/certs \
+    -v $LOG_PATH:/app/setup.log \
+    wikibase/deploy-setup-webserver"
+
+  echo
```
rti, line 90: Is this required? Or is this a dev utility?

deploy/setup/setup.sh
```diff
@@ -0,0 +1,154 @@
...
+    log_cmd "pacman -Sy --noconfirm git"
+  else
+    echo "\u26a0\ufe0f Unsupported package manager. Please install Git manually."
+    exit 1
+  fi
+}
+
+install_docker() {
+  echo "Installing Docker..."
+  log_cmd "curl -fsSL https://get.docker.com | sh"
```
rti, line 100: I would definitely prefer to use docker from the official distributions directories here. Have you considered this? Otherwise in my opinion it is quite intrusive to change settings on the users package manager such as adding repos, adding trusted keys etc.

deploy/setup/setup.sh
```diff
@@ -0,0 +1,154 @@
...
+SETUP_DIR="${SETUP_DIR:-$DEPLOY_DIR/setup}"
+SCRIPTS_DIR="$SETUP_DIR/scripts"
+LOG_PATH="${LOG_PATH:-/tmp/wbs-deploy-setup.log}"
+
+# -- Setup logging --
+
+mkdir -p "$(dirname "$LOG_PATH")"
+touch "$LOG_PATH" 2>/dev/null || true
+
+log() {
```
rti, line 60: Have you considered reusing you `_logging.sh` script?

rti (commented):
Nice work, just did the first round of review.

deploy/setup/setup.sh
lorenjohnson: This omission is on purpose and unfortunately necessary as this script is designed to be ran independently and before having the entire repo cloned. I will add a comment here to this effect, but basically the library scripts/_logging.sh code and this bit need to be manually kept in sync.

deploy/setup/package.json
lorenjohnson: For now this is `private: true` so not managed by nx. But for several reasons I do think this packaged would be versioned separate from the Deploy package. As highly related products they technically actually have very different concerns and are minimally dependent on each other. In the end deploy/setup's interface with that product ends with the generation of a `.env` file and the first run of a `docker compose up -d`.

deploy/setup/setup.sh
lorenjohnson: Sounds worth looking into. I am not wedded to Docker's convenance script here. I am thinking doing some version of what I am doing above with install_git is in order. I choose for now to go this way as my working reference was Dokploy's own convenance script which *if I remember right*, they used this method. What I do know and like about the convenance script is that I have never seen it fail, and I have ran it across several Linux distros, and it is rather quick. But yes maybe we move to the package managers... Leaving open to loop back to later.

deploy/setup/scripts/web-config.sh
lorenjohnson: This is the web-config.sh which creates and boots the Config / Setup page, so it is required if you want to complete setup via Web. There is not an option (which is the default) to just complete the configuration via command line prompts, and in that case this script never gets ran. But was this what you were asking? I wasn't clear on this one.

rti: I think my comments are off by some lines, maybe because files changed while I was writing my comment. I was explicitly asking about the `/app/views` volume because the views get built in in the Dockerfile already.

lorenjohnson: Ah ha, yes that makes sense. The container packaged index.eta is a bit of a fallback, and
the one volume mapped over it is to easy development. I prefer to keep both to keep the image capable as backed, but still a live
editable page for dev. I could make that a dev-only feature but I don’t see much difference or benefit to doing so. Lemme know if you see something else.

deploy/README.md
```diff
@@ -42,11 +42,30 @@ You need two DNS records that resolve to your machine's IP address:
...
+
+### Option 1: Web-based Setup (Recommended for New Users)
+
+If you're setting up a new instance on a fresh VPS, you can use our web-based installer to simplify the process. This method installs all required dependencies, configures your environment, and launches the full Wikibase Suite via a browser-based setup wizard.
+
+1. **Log in to your VPS as root.**
+2. **Run the following one-liner:**
+
+   ```sh
+   bash -c "$(curl -fsSL https://raw.githubusercontent.com/wmde/wikibase-release-pipeline/refs/heads/deploy-setup-script/deploy/setup/setup.sh)"
```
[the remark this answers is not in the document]
lorenjohnson: This article was fairly easy reach, but generally reflects my feelings on this topics :)  https://medium.com/@ewindisch/curl-bash-a-victimless-crime-d6676eb607c9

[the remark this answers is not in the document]
rti: Thanks, nice read. Good points in there.

Co-incidentally, a similar topic just popped up on hacker news yesterday: https://news.ycombinator.com/item?id=44669998

I think my immediate concerns might actually boil down to:
 - incomplete script execution due to aborted download: also mentioned in your link, could be handled by wrapping the execution in a function and starting the function at the very end of the script
 - installing software on the users machines in unexpected/unclean ways: thanks a lot for your efforts in `install_docker()` ❤️

deploy/setup/package.json
rti: I get the theoretical idea. I still do not see what this version communicates to who and when it is updated by who based on which rules.

What I want to prevent is a semi-hidden version number which looks like communicating something, but in the end is just never updated.

deploy/setup/scripts/web-config.sh
rti: Interesting approach 👍🏾

deploy/setup/scripts/web-config.sh
```diff
@@ -0,0 +1,103 @@
...
+  log_cmd "mkdir -p $SETUP_DIR/letsencrypt $CERTS_DIR"
+  log "Using domain: $SETUP_HOST"
+
+  log_cmd "docker pull certbot/certbot"
+
+  if log_cmd "docker run --rm \
+    -v $SETUP_DIR/letsencrypt:/etc/letsencrypt \
+    -v $CERTS_DIR:/certs \
+    -p 80:80 \
+    certbot/certbot certonly \
```
rti, line 52: Do we want to lock the version here? WDYT?

deploy/setup/setup.sh
```diff
@@ -0,0 +1,199 @@
...
+    fedora)
+      log "Detected OS: Fedora"
+
+      log_cmd "dnf install -y dnf-plugins-core"
+      log_cmd "dnf config-manager --add-repo https://download.docker.com/linux/fedora/docker-ce.repo"
+      log_cmd "dnf install -y docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin"
+      log_cmd "systemctl enable --now docker"
+      ;;
+
+    centos)
```
rti, line 135: My initial research revealed no centos docker/compose packages, only podman. Maybe we need to stick with the additional repository on centos. :(

deploy/README.md
```diff
@@ -42,11 +42,30 @@ You need two DNS records that resolve to your machine's IP address:
...
+
+Once setup is complete, the installer shuts down automatically and your Wikibase Suite instance will be up and running.
+
+> 💡 This method takes care of installing Docker, Docker Compose, Git, and configuring everything correctly, making it ideal for first-time deployments.
 
 > 💡 If you want to run a quick test on a machine that has no public IP address (such as your local machine), check our [FAQ entry](#can-i-host-wbs-deploy-locally) below.
 
-### Download WBS Deploy
+### Option 2: Manually Setup
+
```
rti, line 67: Maybe we can move the software requirements section in here, because it is automatic for those using the web based setup

rti:
Canceling the installer during "admin email entry" and running it again leads to the following message:
```
File '/etc/apt/keyrings/docker.gpg' exists. Overwrite? (y/N)
```

deploy/README.md
rti, line 55: This is intended to call the web installer, right? I think we need to add the `--web` arg, atm it runs the cli... which is also nice btw :)

deploy/setup/setup.sh
```diff
@@ -0,0 +1,199 @@
...
+  fi
+}
+
+install_docker() {
+  echo "Installing Docker..."
+
+  # shellcheck disable=SC1091
+  . /etc/os-release
+  case "$ID" in
+    ubuntu|debian)
```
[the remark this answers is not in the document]
lorenjohnson: Given this, I'd like to add better dist + version sensing here, and make a fallback to the original less tidy/secure Docker.io convenance script in cases we know the dist isn't otherwise available or working.

deploy/setup/setup.sh
lorenjohnson: By the additional repo, do you mean what the convenience script sets up? If so, then I agree. I would prefer to resort to the Docker provided convenance script over the two alternatives I see: 1) not supporting the dist+version, or, 2) creating and maintaining our own dist+version specific steps. Given that, the question which remains for me is just how much of a notice/warning/fuss is appropriate and helpful to give to the user in these cases.

deploy/README.md
lorenjohnson: Yes, I am inviting this question now that there is a CLI installer. I am not at all sure that the CLI installer shouldn't be the default, in fact I think for now the CLI installer probably should be. The case for the Web installer would be when initiated through Cloud Config / Init where there is no CLI level user interaction and the user waits until the Web setup becomes available on port 8888. This case is not something I've currently been talking about as I haven't in my own testing found it to be necessarily as simply to explain / guide a potentially inexperienced devops user through the steps to doing. So far it has felt more VPS platform agnostic to forward the "login as root on a fresh VPS instance and run this curl+script script" option as the easiest to not get wrong option.

lorenjohnson:
> > In Progress
> 
> Is not considered scope of this PR, correct?

Yes, sorry for the confusion there. As I need some sort of localhost option for dev'ing this feature itself, there are some localhost considerations and the `--local` CLI arg available in this version, but the `mkcert` cert authority generation and a couple other localhost considerations are indeed to be left out-of-scope of this PR.

lorenjohnson:
@rti the TODO list in the description is now moved to the top of the PR and up-to-date with discussions as of right now.

deploy/setup/setup.sh
rti: Like the approach.

> how much of a notice/warning/fuss is appropriate and helpful to give to the user in these cases

Hm, maybe a note that we are falling back to https://get.docker.com is nice...

lorenjohnson:
@rti extensive refactoring for clarity and testing across all noted distros/versions. A lot of refinements in the UI. Ready for re-review!

deploy/setup/scripts/web-config.sh
lorenjohnson: Yes, and I've went ahead and pinned it now to v4.2.0 (they don't have sematic version series pins like us :)

deploy/setup/package.json
rti: I can see all your points. Still, leaving a `1.0.0` here without a way set how to continue does not feel good to me. Especially because all the other versions in this repo have their clear definitions and processes.

What about setting the version to 0.0.0 for now? It at least declares the version kind of obviously not handled yet, and move on to higher numbers as soon as the NX integration is there. Alternatively, I would be also up for adding the NX integration now.

I am not a fan of the current state "in between".

deploy/setup/setup.sh
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Clear screen
+clear || printf "\033c"
```
rti, line 5: In my terminal this deletes the whole scroll back buffer. I would prefer the script not doing that. I think it is also fine to just start with printing into the terminal in the state it currently is.

deploy/setup/scripts/launch.sh
```diff
@@ -0,0 +1,114 @@
...
+# --- Expected Variables ---
+
+export LOCALHOST
+export DEBUG
+export LOG_PATH
+export DEPLOY_DIR
+export SCRIPTS_DIR
+export SETUP_DIR
+export SKIP_DEPENDENCY_INSTALLS
+export SKIP_LAUNCH
```
rti, line 13: what is the intention of exporting all these here?

deploy/setup/scripts/launch.sh
```diff
@@ -0,0 +1,114 @@
...
+
+# --- New Variables ---
+
+# Internal re-entry flag used when ran detached
+LAUNCH_ONLY=$([ "${1-}" = "--launch-only" ] && echo true || echo false)
```
rti, line 18: Is the comment still up to date? This does not look like a re-entry flag, does it?

deploy/setup/scripts/launch.sh
```diff
@@ -0,0 +1,114 @@
...
+final_message() {
+  {
+    echo
+    echo "✅ Setup is Complete!"
+    echo
+    if [[ -f "$ENV_FILE_PATH" ]]; then
+      while IFS= read -r line; do
+        [[ "$line" =~ ^#.*$ || -z "$line" ]] && continue
+        eval "export $line"
+      done < "$ENV_FILE_PATH"
```
rti, line 50: Is this equal so `source .env`?

deploy/setup/scripts/launch.sh
rti: I see now. Interesting flow. Not obvious on the first sight, but makes sense. Did you consider splitting this into two script files?

lorenjohnson: I will consider it further.

deploy/setup/scripts/install-docker.sh
```diff
@@ -0,0 +1,61 @@
...
+  if [ -z "$CODENAME" ]; then
+    echo "❌ Could not determine distro codename" >&2
+    exit 1
+  fi
+
+  run "echo 'Using Docker APT repo for $OS_ID ($CODENAME)'"
+  run "sudo apt-get update -y"
+  run "sudo apt-get install -y --no-install-recommends ca-certificates curl gnupg"
+  run "sudo install -m 0755 -d /etc/apt/keyrings"
+  run "curl -fsSL https://download.docker.com/linux/$OS_ID/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg"
```
rti, line 33: I'd prefer to use `docker.io` from the ubuntu / debian repos. Is there a reason to not?

lorenjohnson: I preferred this as I was sure to get the more recent versions and the same code runs regardless of dist/version.

deploy/setup/scripts/install-docker.sh
```diff
@@ -0,0 +1,61 @@
...
+  if [ -z "$CODENAME" ] && command -v lsb_release >/dev/null 2>&1; then
+    CODENAME="$(lsb_release -cs 2>/dev/null || true)"
+  fi
+  if [ -z "$CODENAME" ]; then
+    echo "❌ Could not determine distro codename" >&2
+    exit 1
+  fi
+
+  run "echo 'Using Docker APT repo for $OS_ID ($CODENAME)'"
+  run "sudo apt-get update -y"
```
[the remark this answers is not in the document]
lorenjohnson: Yes, it was on my own list and forgot to update. I was trying to re-used code in published recommendations from Docker so some of that crept it. Removed now and re-testing.

deploy/setup/setup.sh
```diff
@@ -0,0 +1,120 @@
...
+  elif command -v yum >/dev/null 2>&1; then
+    # CentOS 7 / Amazon Linux 2 / older RHEL-family (yum-based)
+    # - Legacy yum path; installs Git from the default repos.
+    #   On older CentOS 7 this will be an older Git version, but usually fine.
+    yum install -y git >/dev/null 2>&1
+  else
+    echo "⚠️ Unsupported package manager. Please install Git manually."
+    exit 1
+  fi
+}
```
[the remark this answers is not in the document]
lorenjohnson: I can defend this more somewhere, but the long and short of it is I am actually ok with that for now.

deploy/setup/scripts/launch.sh
```diff
@@ -44,10 +44,8 @@ final_message() {
     echo "✅ Setup is Complete!"
     echo
     if [[ -f "$ENV_FILE_PATH" ]]; then
-      while IFS= read -r line; do
-        [[ "$line" =~ ^#.*$ || -z "$line" ]] && continue
-        eval "export $line"
-      done < "$ENV_FILE_PATH"
+      # shellcheck disable=SC1090
+      source "$ENV_FILE_PATH"
```
rti, line 48: you might need to set the `allexport` option too
https://stackoverflow.com/a/30969768/156191

deploy/setup/scripts/install-docker.sh
rti: I have the strong opinion that it is not good practice to get packages from external sources if distro repos provide what is needed. Is `docker.io` too old in any of the distro repos?

lorenjohnson: Yes, my understanding from you and from testing was that at least one of the Debian repos for one of the supported Debian versions was out of date (only providing docker 1.x)

deploy/setup/setup.sh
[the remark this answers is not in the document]
lorenjohnson: Did you have a recommendation here in terms of code?

deploy/setup/scripts/web-config.sh
```diff
@@ -0,0 +1,114 @@
...
+    echo "⚠️ This setup page is using a temporary self-signed HTTPS certificate."
+    echo "Your browser will likely show a warning before loading the page."
+    echo "See the Troubleshooting section of the Deploy Setup README for help"
+    echo "if you want to bypass the warning or replace it with a trusted cert."
+    echo
+  fi
+  if $LOCALHOST; then
+    echo  It is now safe to close this terminal session.
+    echo
+  fi
```
rti, line 107: Why is this only safe in the localhost case?

deploy/README.md
rti: I think I personally really prefer the reordering as initially suggested.

deploy/setup/setup.sh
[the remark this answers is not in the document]
rti: My bad. I think I misread the piping going on there. If the stderr output is in the logfile, I am happy. Thanks for clarifying.

deploy/setup/scripts/install-docker.sh
rti: Yes, docker-compose was the problem on debian. Now with trixie being stable, we also have compose v2 natively on debian https://packages.debian.org/search?suite=trixie&arch=any&searchon=names&keywords=docker-compose

PR #883, opened 2025-07-21
feat: depth 1 checkouts in wikibase-image build

rti:
Seems to stabilize image builds. Sometimes gerrit would just drop the connection in the middle of the checkout otherwise.


lorenjohnson (approved):
LGTM. I appreciate the extra round of code organization and clarifying done here too.

PR #885, opened 2025-07-22
T399694 - Enables users to run WBS via Dokploy or other PaaS providers

lorenjohnson:
Related: https://phabricator.wikimedia.org/T399694#11023565

This work is complete and ready for review, but requires product review. Its goal is to make WBS compatible with platforms such as Dokploy and Coolify, which provide their own reverse proxy.

The options are:

- Use `docker-compose.traefik.yml`, leaving `docker-compose.override.yml` available for user customizations. This changes the documented startup command from `docker compose up` to `docker compose -f docker-compose.yml -f docker-compose.traefik.yml up`.
- Use `docker-compose.override.yml` for Traefik and provide an ignored `docker-compose.local.yml` for user customizations. Users who need it must specify both files explicitly.

This PR takes the second approach, preserving the established `docker compose up` workflow.


rti (approved):
Neat way of splitting up the compose while still keeping the docker compose command line interface the same. Nice addition! Thanks

PR #891, opened 2025-08-25
 T399704 -- feat(deploy): add support for local docker-compose overrides

lorenjohnson:
Adds support for local overrides in Deploy by introducing a deploy/.gitignore (with docker-compose.local.yml ignored). This allows the dev team to test local Docker image changes without modifying tracked files, while keeping Deploy upgradeable for users.

https://phabricator.wikimedia.org/T399704


rti (approved):
Sounds like a desirable addition. Thanks!

PR #893, opened 2025-08-25
fix: bump wikbase container to 5

rti:


rti:
@lorenjohnson if you agree with that procedure, I would just go ahead, merge and release.

PR #895, opened 2025-09-04
style: change default title to "Wikibase Query Service"

rti:
Change requested by PM


lorenjohnson (approved):
Not sure if you wanted a review on this, but if so I approve. FWIW I see this as fine as either a minor or patch update, and needn't wait for a major release.

PR #923, opened 2026-04-16
feat: Include wikibaseType field in callback payload

JacobThwaites:
As a follow up to [this change](https://github.com/wmde/wikibase-metadata/pull/147) on wikibase-metadata, this PR sets the `wikibaseType` field in the callback payload. This ensures that entries are categorized correctly in the metrics.

### DIFF
--- a/build/wikibase/CHANGELOG.md
+++ b/build/wikibase/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 6.0.1 (2026-04-21)
+
+### 🩹 Fixes
+- Include the wikibaseType field inside the callback payload
+
 # 6.0.0 (2026-02-16)
 
 Upgrades to MediaWiki from 1.44 to 1.45 including updates to compatible packaged extensions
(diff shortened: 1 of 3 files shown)


RickiJay-WMDE (approved):

lorenjohnson:
Hi @JacobThwaites,

- Please look into the SHELLCHECK linting issues before merging. Shell files only get linted when they are part of a changeset, and unfortunately it appears you're inheriting linting issues that were unresolved on the initial implementation of the callback script.  See https://www.shellcheck.net and it is ok by me to ignore rules here if it is awkward to comply. See https://www.shellcheck.net for more, and let me know if you need a hand with things there. 

- Make sure when merging that either the merge commit message or the enclosed commit message(s) are in conventional commits format (ideally both). I would recommend "fix(wikibase): include wikibaseType in callback payload” (see https://github.com/wmde/wikibase-release-pipeline/blob/main/docs/conventional-commits.md)

PR #924, opened 2026-04-20
feat(wikibase): disable anonymous writes by default

lorenjohnson:
In parity with Wikibase Cloud defaults, this updates the Wikibase image defaults so anonymous users keep read access, but can no longer:

- edit pages
- create pages
- create talk pages
- use the write API
- create accounts

Notes:

- Our test suite relied heavily on anonymous write access, so part of this changeset is updates to test suites that now require a logged in user. Fortunately, most of the required changes were able to be carried over from the now rejected Temporary Accounts work that had introduced a similar requirement on our tests. 
- I have also moved the image-managed `LocalSettings` loading into dedicated files so the generated `LocalSettings.php` keeps stable DO NOT REMOVE `require_once` lines while still leaving a supported place for user-defined settings before and after bundled extensions load. This ensures the new defaults are attributes of the image, and come with the upgrade vs the user needing to replace their generated `LocalSettings.php`.

A more comprehensive description of this change and related decision is in the included ADR: [docs/adr/0018-anonymous-access-defaults.md](https://github.com/wmde/wikibase-release-pipeline/blob/22c55bb01b732f2e23e86013d7c54aa338748ddc/docs/adr/0018-anonymous-access-defaults.md)

### DIFF
--- a/build/wikibase/Dockerfile
+++ b/build/wikibase/Dockerfile
@@ -225,6 +225,8 @@ COPY callback.sh /callback.sh
 COPY jobrunner-entrypoint.sh /jobrunner-entrypoint.sh
 COPY htaccess /var/www/html/.htaccess
 COPY LocalSettings.d LocalSettings.d
+COPY LocalSettings.MediaWiki.php /LocalSettings.MediaWiki.php
+COPY LocalSettings.Extensions.php /LocalSettings.Extensions.php
 COPY default-extra-install.sh /default-extra-install.sh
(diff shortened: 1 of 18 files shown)


JacobThwaites (approved):

lorenjohnson:
Thanks for the review @JacobThwaites! To answer the passing comment that I think you may have retracted, that comment was ferried across from the earlier location of that file, so kept for now, and not related to current changes. But a bit musty for sure.

PR #927, opened 2026-04-23
fix(wikibase): fix callback enum payload and clarify mul changelog note

lorenjohnson:
- Fixes the Wikibase callback payload from #923 such that `wikibaseType` is sent as the GraphQL enum `SUITE` instead of the quoted string `"SUITE"`. This callback regression caused the metadata registration mutation to fail with `Enum 'WikibaseType' cannot represent non-enum value: "SUITE"` during Wikibase startup, and would cause Wikibase service to fail loading for any users opting in to Metadata reporting. 
- Clarifies the 7.0.0 Wikibase image CHANGELOG note for `mul` support so the Elasticsearch reindex requirement is better highlighted


JacobThwaites:
Looks good! This resolves a mismatch between the data-type for the `wikibaseType` field expected from wikibase-metadata and what is actually sent.

JacobThwaites (approved):

PR #954, opened 2026-07-31
test: split Query Service suite

lorenjohnson:
Moves the Query Service tests into a dedicated suite and uses a test-only Docker Compose profile so WDQS, its updater, and frontend only start for that suite. The published root Docker Compose configuration remains unchanged.


lorenjohnson:
Moved this commit directly onto test-improvements as requested; validation will continue on PR #953.

